PR #40, opened 2022-06-09
Allow client port to be specified

ben476:
WebTransport can be constructed with a prop `localPort` to specify what port the client should bind to. It's useful when doing stuff like UDP hole punching.

### DIFF
--- a/src/http3client.cc
+++ b/src/http3client.cc
@@ -83,10 +83,11 @@ namespace quic
                              QuicSocketAddress server_address, const std::string &server_hostname,
                              std::unique_ptr<ProofVerifier> proof_verifier,
                              std::unique_ptr<SessionCache> session_cache,
-                             std::unique_ptr<QuicConnectionHelperInterface> helper)
+                             std::unique_ptr<QuicConnectionHelperInterface> helper,
+                             int local_port)
         : server_id_(QuicServerId(server_hostname, server_address.port(), false)),
           initialized_(false),
-          local_port_(0),
+          local_port_(local_port),
           store_response_(false),
           latest_response_code_(-1),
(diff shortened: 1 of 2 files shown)


martenrichter:
I am a bit puzzled. Does the local port differ from the port that is used to do the UDP communication?
I have looked into the code and I have already the port property, but it seems, that I forgot to pass it to Http3Client object?
Is this the same? Then it is a bug

martenrichter:
I looked into it again; it seems to be the port on which the client is listening for the UDP replies.
Ok, that is something different. Is it already included in the Webtransport spec https://www.w3.org/TR/webtransport/#webtransport-constructor
I do not see it there...

I want clarification on what it does, and I have only 10 minutes in the morning today. On the weekend I have more time.

ben476:
Whoops, I didn't realise we're supposed to follow the spec. 

Indeed, you're correct. What I'm proposing is the ability for the client to specify what port to bind to instead of having the OS pick a random free port. This isn't in the spec probably for security reasons in a browser context, but I don't think it'll be an issue in node.js.

I think this would be a handy feature if there's room for flexibility because it's not running in the browser. As an example use case, I'm trying to get WebTransport to work peer to peer, which needs the client to use a pre-negotiated port (like WebRTC).

martenrichter:
Well let's put it this way, I try to follow the spec as close as possible. Since I want to allow that almost the same code can use this lib in node.js or the normal browser. That is why I am asking, so I always check, if there is something in the spec.
But I also see no problem in adding this in the node.js case.
(May be you should consider suggesting this change also to the authors of the spec?)
So I will approve the merge, new release is coming later

Issue #93, opened 2022-10-26
Add types?


achingbrain:
Would you accept a PR that adds types to the js part of this repo?

Either a) generated from JSDocs or b) conversion to typescript.

a) normally sounds like it'd be a smaller changeset but usually ends up more painful due to needing to unpick type conversions, null guarding on instance fields that aren't set in constructors etc and having to add `/** @type {foo} */` in loads of places.  b) is usually easier to work with going forwards and lets you express certain things more clearly (public interfaces, inheritance, etc).

martenrichter:
I am generally open to including the types since it will make the life for people using TS easier.
Regarding a) or b), so far I have not looked at typescript, so I am not familiar with it.
On the other hand, the interface should be stable, since it tries to match the spec.
And of course, the types should match the current webtransport spec.

What I am not sure, what happens to the napi side,  types have side effects on this part. probably not?

achingbrain:
> What I am not sure, what happens to the napi side

One approach is to [hand write the interfaces the native modules expose](https://stackoverflow.com/questions/44275172/typescript-declarations-file-for-node-c-addon) - a bit error prone but it's better than nothing.

martenrichter:
> > What I am not sure, what happens to the napi side
> 
> One approach is to [hand write the interfaces the native modules expose](https://stackoverflow.com/questions/44275172/typescript-declarations-file-for-node-c-addon) - a bit error prone but it's better than nothing.

Or can one ignore this part, since the napi interfaces are never exposed to the outside world?

achingbrain:
It's very tempting to say only the outside interface needs types, but `tsc` will verify the types are accurate - values returned have the correct fields and so on, which means they start to be applied internally as well.

At this point people usually decide to hand-roll the external types instead, which are rarely accurate and even if they are, tend to start diverging immediately.  The alternative is to mark everything internally as the `any` type which gives you absolutely nothing in terms of type safety and just adds noise everywhere.

For the c++ part, I'm not sure, something extracted from [nbind](https://www.npmjs.com/package/nbind) could help but I think this is less urgent since it's 100% internal.

I've typed the module using JSDoc locally, I'm just sanity checking it all against my use of the module in another project, when that's done I'll open a PR.

martenrichter:
[...]

PR #94, opened 2022-10-27
feat: add types to module

achingbrain:
* Generate types from JSDoc
* Adds `src/dom.ts` - types take from the WebTransport spec
* Adds `src/types.ts` - interfaces used by `src/webtransport.js`
* Adds types to `src/webtransport.js` and all tests
* Exposes WebTransport spec types to hide internal implementation details

Some things to consider:

* Adds functions and properties to conform to the spec more closely, mostly around stats
* Where these stats are not available they return `0` or `0n` - I figure these can be filled in at a later date
* `number` is used for `DOMHighResTimeStamp` as it is a `double`
* `bigint` is used for `unsigned long long` as it can overflow a `number`
* I can't see any of the `unsigned long long` fields used in Chrome's `WebTransport` implementation so I'm unsure if `bigint` is correct or if `number` will do, we'll probably have to wait and see what's implemented

I performed some small refactorings:

* `Http3Client` has had it's handler methods split out of `customCallback` as it seemed consistent with the other classes
* `Http3EventLoop.globalLoop` is created as a local variable otherwise you have to null guard on every `Http3EventLoop.globalLoop` access
* In order to type instance fields it's sometimes necessary to set them in the constructor
* In some places it was necessary to add additional null-guarding

Improvements we may wish to make:

* The `webtransport.js` file is quite large, splitting the classes out into separate files would aid readability and testing
* The whole of the `/src` dir is bundled into the release, consider moving `.js` into `/lib` and keeping `/src` for `.cc`/`.h` files and only including `/lib`
* Does the `platform` dir need to be included in the release too?
* A lot of code could be simplified by reusing other modules - `p-defer` instead of having `this.fooResolve` and `this.fooReject` everywhere for example
* Error handling seems quite limited - for example waiting for a client session to be ready hangs forever if there's an error connecting
* `console.log` is used in the non-test code which makes controlling output hard
* There are random delays in places, starting a server then setting a timeout before connecting to it, for example - better to use events like `"listening"` same as the node core `http` module

Fixes: #93

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,14 +8,15 @@
     "node": ">=16"
   },
   "type": "module",
+  "types": "./dist/src/webtransport.d.ts",
   "scripts": {
     "start": "node test/echoserver.js",
     "test": "node test/test.js",
-    "install": "node build.js install",
-    "build": "node build.js build",
-    "rebuild": "node build.js rebuild",
-    "rebuild-debug": "node build.js rebuild-debug -D",
-    "build-debug": "node build.js build-debug"
+    "install": "node build.js install && tsc",
+    "build": "node build.js build && tsc",
+    "rebuild": "node build.js rebuild && tsc",
+    "rebuild-debug": "node build.js rebuild-debug -D && tsc",
+    "build-debug": "node build.js build-debug && tsc"
   },
   "repository": {
     "type": "git",
@@ -38,6 +39,7 @@
   "devDependencies": {
     "@babel/core": "^7.17.10",
     "@babel/eslint-parser": "^7.17.0",
+    "@types/node": "^18.11.5",
     "cmake-js": "^6.3.2",
     "eslint": "^7.32.0",
     "eslint-config-prettier": "^7.2.0",
@@ -48,7 +50,8 @@
     "eslint-plugin-promise": "^4.3.1",
     "eslint-plugin-standard": "^4.1.0",
     "node-forge": "^1.3.1",
-    "prettier": "^2.4.1"
+    "prettier": "^2.4.1",
+    "typescript": "^4.8.4"
   },
   "binary": {
     "napi_versions": [
--- a/package.json
+++ b/package.json
@@ -12,11 +12,12 @@
   "scripts": {
     "start": "node test/echoserver.js",
     "test": "node test/test.js",
-    "install": "node build.js install && tsc",
-    "build": "node build.js build && tsc",
-    "rebuild": "node build.js rebuild && tsc",
-    "rebuild-debug": "node build.js rebuild-debug -D && tsc",
-    "build-debug": "node build.js build-debug && tsc"
+    "install": "node build.js install",
+    "build": "node build.js build",
+    "rebuild": "node build.js rebuild",
+    "rebuild-debug": "node build.js rebuild-debug -D",
+    "build-debug": "node build.js build-debug",
+    "types": "tsc"
   },
   "repository": {
     "type": "git",
--- a/src/dom.ts
+++ b/src/dom.ts
@@ -31,11 +31,11 @@ export interface WebTransportCloseInfo {
 export interface WebTransportDatagramDuplexStream {
   readable: ReadableStream<Uint8Array>
   writable: WritableStream<Uint8Array>
-  readonly maxDatagramSize: number
-  incomingMaxAge?: number
-  outgoingMaxAge?: number
-  incomingHighWaterMark: number
-  outgoingHighWaterMark: number
+  // readonly maxDatagramSize: number
+  // incomingMaxAge?: number
+  // outgoingMaxAge?: number
+  // incomingHighWaterMark: number
+  // outgoingHighWaterMark: number
 }
 
 export interface WebTransportBidirectionalStream {
@@ -76,9 +76,9 @@ export interface WebTransportOptions {
 }
 
 export interface WebTransport {
-  getStats: () => Promise<WebTransportStats>
+  // getStats: () => Promise<WebTransportStats>
   readonly ready: Promise<void>
-  readonly reliability: WebTransportReliabilityMode
+  // readonly reliability: WebTransportReliabilityMode
   readonly closed: Promise<WebTransportCloseInfo>
   close: (closeInfo?: WebTransportCloseInfo) => void
   readonly datagrams: WebTransportDatagramDuplexStream
--- a/src/dom.ts
+++ b/src/dom.ts
@@ -1,0 +1,95 @@
+// Types taken from https://www.w3.org/TR/webtransport
+// These can be removed when they are added to the default typescript types
+
+export interface WebTransportDatagramStats {
+  timestamp: number
+  expiredOutgoing: bigint
+  droppedIncoming: bigint
+  lostOutgoing: bigint
+}
+
+export interface WebTransportStats {
+  timestamp: number
+  bytesSent: bigint
+  packetsSent: bigint
+  packetsLost: bigint
+  numOutgoingStreamsCreated: number
+  numIncomingStreamsCreated: number
+  bytesReceived: bigint
+  packetsReceived: bigint
+  smoothedRtt: number
+  rttVariation: number
+  minRtt: number
+  datagrams: WebTransportDatagramStats
+}
+
+export interface WebTransportCloseInfo {
+  closeCode: number
+  reason: string
+}
+
+export interface WebTransportDatagramDuplexStream {
+  readable: ReadableStream<Uint8Array>
+  writable: WritableStream<Uint8Array>
+  readonly maxDatagramSize: number
+  incomingMaxAge?: number
+  outgoingMaxAge?: number
+  incomingHighWaterMark: number
+  outgoingHighWaterMark: number
+}
+
+export interface WebTransportBidirectionalStream {
+  readonly readable: ReadableStream<Uint8Array>
+  readonly writable: WritableStream<Uint8Array>
+}
+
+export interface  WebTransportSendStreamStats {
+  timestamp: number
+  bytesWritten: bigint
+  bytesSent: bigint
+  bytesAcknowledged: bigint
+}
+
+export interface WebTransportSendStream extends WritableStream<Uint8Array> {
+  getStats: () => Promise<WebTransportSendStreamStats>
+}
+
+export interface WebTransportReceiveStreamStats {
+  timestamp: number
+  bytesReceived: bigint
+  bytesRead: bigint
+}
+
+export interface WebTransportReceiveStream extends ReadableStream<Uint8Array> {
+  getStats: () => Promise<WebTransportReceiveStreamStats>
+}
+
+export interface WebTransportHash {
+  algorithm: string
+  value: BufferSource
+}
+
+export interface WebTransportOptions {
+  allowPooling?: boolean
+  requireUnreliable?: boolean
+  serverCertificateHashes?: WebTransportHash[]
+}
+
+export interface WebTransport {
+  getStats: () => Promise<WebTransportStats>
+  readonly ready: Promise<void>
+  readonly reliability: WebTransportReliabilityMode
+  readonly closed: Promise<WebTransportCloseInfo>
+  close: (closeInfo?: WebTransportCloseInfo) => void
+  readonly datagrams: WebTransportDatagramDuplexStream
+
+  createBidirectionalStream: () => Promise<WebTransportBidirectionalStream>
+  /* a ReadableStream of WebTransportBidirectionalStream objects */
+  readonly incomingBidirectionalStreams: ReadableStream<WebTransportBidirectionalStream>
+
+  createUnidirectionalStream: () => Promise<WebTransportSendStream>
+  /* a ReadableStream of WebTransportReceiveStream objects */
+  readonly incomingUnidirectionalStreams: ReadableStream<WebTransportReceiveStream>
+}
+
+export type WebTransportReliabilityMode = 'pending' | 'reliable-only' | 'supports-unreliable'
--- a/src/types.ts
+++ b/src/types.ts
@@ -1,0 +1,133 @@
+import type { WebTransport } from './dom'
+
+/**
+ * Native Http3WTSession counterpart
+ */
+ export interface NativeHttp3WTSession {
+  jsobj: WebTransportSessionEventHandler
+  writeDatagram: (chunk: Uint8Array) => void
+  orderUnidiStream: () => void
+  orderBidiStream: () => void
+  close: (arg: { code: number, reason: string }) => void
+}
+
+/**
+ * Native Http3WTStream counterpart
+ */
+export interface NativeHttp3WTStream {
+  jsobj: WebTransportStreamEventHandler
+  startReading: () => void
+  stopReading: () => void
+  stopSending: (code: number) => void
+  resetStream: (code: number) => void
+  writeChunk: (buf: Uint8Array) => void
+  streamFinal: () => void
+}
+
+export type Purpose = 'StreamRecvSignal' | 'StreamRead' | 'StreamWrite' | 'StreamReset' | 'StreamNetworkFinish'
+export type NetTask = 'stopSending' | 'resetStream' | 'streamFinal'
+
+export interface StreamRecvSignalEvent {
+  object: NativeHttp3WTStream
+  purpose: 'StreamRecvSignal'
+  code: number
+  nettask: NetTask
+}
+
+export interface StreamReadEvent {
+  object: NativeHttp3WTStream
+  purpose: 'StreamRead'
+  data?: Uint8Array
+  fin?: boolean
+}
+
+export interface StreamWriteEvent {
+  object: NativeHttp3WTStream
+  purpose: 'StreamWrite'
+}
+
+export interface StreamResetEvent {
+  object: NativeHttp3WTStream
+  purpose: 'StreamReset'
+}
+
+export interface StreamNetworkFinishEvent {
+  object: NativeHttp3WTStream
+  purpose: 'StreamNetworkFinish'
+  nettask: NetTask
+}
+
+export interface WebTransportStreamEventHandler {
+  onStreamRecvSignal: (evt: StreamRecvSignalEvent) => void
+  onStreamRead: (evt: StreamReadEvent) => void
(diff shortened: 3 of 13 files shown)


src/webtransport.js
```diff
@@ -606,9 +858,13 @@ export class Http3Server extends Http3WebTransport {
...
   }
 
+  /**
+   * @param {string} path
+   * @returns {ReadableStream<WebTransportSession>}
+   */
   sessionStream(path) {
     if (path in this.sessionStreams) {
-      return this.sessionsStreams[path]
+      return this.sessionStreams[path]
```
achingbrain, line 867: This appears to be a typo, it would likely explode if invoked at runtime.

martenrichter: I think it is a typo. Interesting that it did not cause harm so far.

martenrichter:
The first is look very nice, I will look in detail at it on the weekend, so more comments will follow.
Can you please:
`Http3EventLoop.globalLoop is created as a local variable otherwise you have to null guard on every Http3EventLoop.globalLoop access`
make initially `null` again, starting the eventloop is quite heavy in terms of memory and CPU, so it should be there on demand. Thanks

martenrichter:
> Some things to consider:
> 
> * Adds functions and properties to conform to the spec more closely, mostly around stats
> * Where these stats are not available they return `0` or `0n` - I figure these can be filled in at a later date
> * `number` is used for `DOMHighResTimeStamp` as it is a `double`
> * `bigint` is used for `unsigned long long` as it can overflow a `number`
> * I can't see any of the `unsigned long long` fields used in Chrome's `WebTransport` implementation so I'm unsure if `bigint` is correct or if `number` will do, we'll probably have to wait and see what's implemented

This object 
`https://github.com/google/quiche/blob/main/quiche/quic/core/quic_connection_stats.h`
should give some information, what information is available from quiche and hints about potential types.

martenrichter:
> * The whole of the `/src` dir is bundled into the release, consider moving `.js` into `/lib` and keeping `/src` for `.cc`/`.h` files and only including `/lib`
> * Does the `platform` dir need to be included in the release too?

All source code from the c++ side is required, if the platform has no prebuild library.
The `platform` dir is for platform implementation of the quiche build.
So we can split but all need to be inside the release bundle.

martenrichter:
> * A lot of code could be simplified by reusing other modules - `p-defer` instead of having `this.fooResolve` and `this.fooReject` everywhere for example

Actually, I am not a fan of including mini modules for simple tasks, since this allows supply chain attacks.

martenrichter:
> * Error handling seems quite limited - for example waiting for a client session to be ready hangs forever if there's an error connecting

That is partially true, that there is room for improvement of c++ code passing on errors. There PRs are welcome, but probably more on the c++ side.

martenrichter:
> * `console.log` is used in the non-test code which makes controlling output hard

These should go away sometime. So far, I leave them, until all errors are gone for a while. They help with diagnosis. However, these should be limited to the event loop. and very few tasks.

martenrichter:
> * There are random delays in places, starting a server then setting a timeout before connecting to it, for example - better to use events like `"listening"` same as the node core `http` module

I am not sure, which events you mean. On the test code? These are not random but carefully adjusted to typical times. Listening to events would not detect if something takes longer than usual or hangs (problems with events loops, so these allowed me in the past to spot problems with the node event loop on the test in CI/CD pipeline.

You mean the C++ side, these are all taken/inspired by the quiche samples, if any are present.

src/webtransport.js
```diff
@@ -168,15 +229,21 @@ class Http3WTStream {
       const res = this.pendingres
       this.pendingoperation = null
       this.pendingres = null
-      res()
```
martenrichter, line 171: Should error, because it indicates a bug in the implementation

src/webtransport.js
```diff
@@ -191,16 +258,24 @@ class Http3WTStream {
...
+  /**
+   * @param {StreamWriteEvent} args
+   */
   onStreamWrite(args) {
     // we ignore success
     if (this.pendingoperation) {
       const res = this.pendingres
       this.pendingoperation = null
       this.pendingres = null
-      res()
```
martenrichter, line 200: Should error, because it indicates a bug in the implementation

src/webtransport.js
```diff
@@ -241,7 +319,9 @@ class Http3WTStream {
           const res = this.pendingres
           this.pendingoperation = null
           this.pendingres = null
-          res()
```
martenrichter, line 244: Should error, because it indicates a bug in the implementation

src/webtransport.js
```diff
@@ -497,42 +669,77 @@ class Http3WTSession {
...
+    if (res != null) {
+      res()
+    }
   }
 
+  /**
+   * @param {SessionReadyEvent | SessionCloseEvent | DatagramReceivedEvent | DatagramSendEvent | NewStreamEvent} args
+   */
   static callback(args) {
+    console.log('Session callback called', args)
```
martenrichter, line 731: Please comment, only for debugging while developing

src/webtransport.js
```diff
@@ -802,10 +1125,42 @@ export class WebTransport {
...
+  /** @returns {import('./dom').WebTransportReliabilityMode} */
+  get reliability () {
+    return 'reliable-only'
+  }
 }
 
 class Http3EventLoop {
-  static globalLoop = null
+  /** @type {Http3EventLoop | null} */
+  static globalLoop = Http3EventLoop.createGlobalEventLoop()
```
martenrichter, line 1159: Needs to be null. Since the object is resource-heavy and it is actually torn down, if no one uses it, and then it is again null.  But null guards may be unnecessary since they have to fail if it is null when it is called. It should not fail silently.

package.json
```diff
@@ -8,14 +8,15 @@
...
+  "types": "./dist/src/webtransport.d.ts",
   "scripts": {
     "start": "node test/echoserver.js",
     "test": "node test/test.js",
-    "install": "node build.js install",
-    "build": "node build.js build",
-    "rebuild": "node build.js rebuild",
-    "rebuild-debug": "node build.js rebuild-debug -D",
-    "build-debug": "node build.js build-debug"
+    "install": "node build.js install && tsc",
```
martenrichter, line 15: Actually, I would prefer if the `tsc` can go into the `build.js`. The build process is quite complex and therefore I need the build.js, furthermore `&&` fails on windows and would not pass the unit tests.

src/dom.ts
```diff
@@ -0,0 +1,95 @@
...
+}
+
+export interface WebTransportHash {
+  algorithm: string
+  value: BufferSource
+}
+
+export interface WebTransportOptions {
+  allowPooling?: boolean
+  requireUnreliable?: boolean
```
martenrichter, line 74: Interesting this a legacy option, http/2 connections are not part of the specs anymore

src/webtransport.js
```diff
@@ -316,50 +416,74 @@ class Http3WTSession {
...
+            }
+            this.objint.writeDatagram(chunk)
+            return ret
+          } else throw new Error('chunk is not of type Uint8Array')
+        },
+        close: () => {
+          // do nothing
+        }
+      }),
+      maxDatagramSize: 0,
```
martenrichter, line 457: Are you sure, that the following properties have suitable default values?

martenrichter:
> * In some places it was necessary to add additional null-guarding

That is probably a good idea, but some of your null guards (at least one), lead to problems.
In some places, I did not include null guards since the object should fail, if it is not set, 
For these cases, if you place a null guard, you have to throw an error to indicate a problem.
For example line 270 or 322 in webtransport.js and others. Maybe this is a bad style on my side...

martenrichter:
Connect issue https://github.com/fails-components/webtransport/issues/93

src/webtransport.js
martenrichter: Or no `this.pendingoperation` guards this already, does `tsc` require an explicit type checking?

src/webtransport.js
martenrichter: Or no `this.pendingoperation` guards this already, does `tsc` require an explicit type checking?

src/webtransport.js
martenrichter: Or no `this.pendingoperation` guards this already, does `tsc` require an explicit type checking?

src/webtransport.js
achingbrain: Ha, oops.  I think I added that for local debugging.

src/webtransport.js
martenrichter: Yes `maxDatagramSize` can be pulled from c++ side. But until it is implemented by someone  0 is a very bad value, null would be better. Since it would say that the datagram should have 0 size.
The other values such as `incomingMaxAge`, should also be left out, since this logic is not implemented, and have to be implemented on the js side.

martenrichter: What about excluding yet not implemented fields from the type definition (commenting for later)? In this way, the compiler will error if the fields are used giving the authors a hint, that the implementation should differ from the browser side.

src/dom.ts
martenrichter: Better remove, comment it.

src/types.ts
```diff
@@ -0,0 +1,133 @@
+import type { WebTransport } from './dom'
+
+/**
+ * Native Http3WTSession counterpart
+ */
+ export interface NativeHttp3WTSession {
```
martenrichter, line 6: Question: this is an internal interface and should not be exported outside of the package, could this be prevented, the same goes for other interfaces that are not part of the spec.

martenrichter: Same goes with internal types

src/webtransport.js
martenrichter: Concerning the watermarks, it is impossible to support these properties also in the future. Since it is not exposed by the normal stream, which we use. So these have to be removed from the interface.

src/webtransport.js
```diff
@@ -555,9 +762,40 @@ class Http3WTSession {
       }
     } else throw new Error('no purpose Sessioncb')
   }
+
```
martenrichter, line 765: Here I am not sure if it should not be commented until it is implemented.

src/dom.ts
```diff
@@ -0,0 +1,95 @@
...
+  serverCertificateHashes?: WebTransportHash[]
+}
+
+export interface WebTransport {
+  getStats: () => Promise<WebTransportStats>
```
martenrichter, line 79: Comment yet not implemented features

src/dom.ts
```diff
@@ -0,0 +1,95 @@
...
+  readonly ready: Promise<void>
+  readonly reliability: WebTransportReliabilityMode
```
martenrichter, line 81: This is legacy, comment, it makes also no sense in the spec

martenrichter (changes requested):
Please look in the comment in the patch and in the conversations.
I really like your changes. But some of them may break some platforms.
Furthermore, the unimplemented parts should be commented. I think it causes more harm pretending a features is there, compared to an immediate error, that signals the developer, that one can not rely on this feature on the node platfrom yet.

src/dom.ts
achingbrain: There's some more context here: https://github.com/w3c/webtransport/issues/107 but the 'reliability' property tells us if we've fallen back to http2/tcp for the connection so this should probably stay in?

martenrichter: Well, they dropped http/2 as well as quic support entirely in favor of http/3. So it was just not deleted.

src/types.ts
achingbrain: When you have a JSDoc at the top of the file like:

```js
/**
 * @typedef {import('./dom'). NativeHttp3WTSession} NativeHttp3WTSession
 */
```

it will also export the type from the current file.

With JSDoc types to prevent this we'd need to split the class that references the internal type from the `@typedef` into it's own file, and import that class from the module's entry point.

With pure typescript you can just not export the interface which keeps it internal.

src/dom.ts
achingbrain: Got a link?  I'm still learning about this stuff..

src/types.ts
martenrichter: Ok, then splitting the file would be a good idea, so that no one tries to hack into it.  Since these interfaces are not guaranteed to stay the same between patch releases. Or do you think it will not cause harm?

src/webtransport.js
achingbrain: > But until it is implemented by someone 0 is a very bad value, null would be better

The spec type is `readonly attribute unsigned long maxDatagramSize`, not `readonly attribute unsigned long maxDatagramSize?` so it cannot be null.

> What about excluding yet not implemented fields from the type definition

I guess this is the key question.  The spec defines all these fields and their types, if we add types that diverge from the spec then we're not following the spec.

I would err on the side of consistency with the spec even though the values may be nonsensical (because one day they will hopefully be sensical), but I'm getting the feeling you'd rather not implement the spec fully?

src/dom.ts
martenrichter: A single link is hard to find...
Here is an old spec 
https://datatracker.ietf.org/doc/html/draft-vvv-webtransport-quic-00
called quictransport and there were many iterations of the draft. Furthermore you see it in the quiche source code, there are still relicts. But you also see that is not capable of http2 webtransport at all. So the best reference is the quiche source code and in the web transport spec the links to the specification of the underlying protocols at the end. See § 10 and its references.

src/types.ts
achingbrain: I would also like the files to be split - I didn't do it as part of this PR as it would make it a lot harder to review..

src/webtransport.js
martenrichter: The spec is for a client in a webbrowser. The server side is fully unspecified. (and yes we have a client as well, but I assume that is not the main application). So there are natural divergences. 
Anyway I try to be as close as possible to the client spec,  so that one can write code, that works on the browser and client in the same way.

And yes I would rather implement the spec not fully. 
Or my approach as follows, I needed the webtransport stuff for a project, that why I wrote. I implemented all that I currently needed. The idea is, that if someone needs a feature, that is currently not implemented, I will help to guide a PR to improve the package functionality.
But I also want clear indications (errors), of what is currently not implemented. I hate it, if I have a lib and after hours of searching I find out, that it errors just, because it pretends to have it.

But if time progresses I hope we will get close and closer to full features in client spec, except for some differences. One is that I use normal web streams, the streams of webtransport are close to webstreams but differ slightly. Therefore I can not implement a completely identical interface.

src/types.ts
martenrichter: ok then we can make it in a follow up PR?

package.json
achingbrain: I've added it to `build.js` and also added `await`s to the invocations of `execbuild` which seemed to be missing.

I've also added a `"types"` npm script that just invokes `tsc` in case you want to build the types without building the C++ code first.

src/webtransport.js
achingbrain: I've removed this.

achingbrain: Or at least, put it back to how it was..

package.json
martenrichter: Looks good..

src/webtransport.js
achingbrain: `tsc` requires explicit type checking - if the value can be `null` it needs be handled as such rather than just letting it fail with a runtime error.

src/types.ts
achingbrain: Yeah, I think that's best, the changes here are disruptive enough!

src/webtransport.js
achingbrain: > but I assume that is not the main application

For my usecase it's both as client and server in node.  Essentially making [@libp2p/webtransport](https://www.npmjs.com/package/@libp2p/webtransport) work in node as well as Chrome.

> And yes I would rather implement the spec not fully.

Ok, I'll remove the stuff added to satisfy the spec, then comment it out of the types and add a note to the README about where this module diverges from the spec.  That should hopefully cover all of the bases?

src/webtransport.js
martenrichter: Perfect, but the second chance in the commit will cause trouble.

src/webtransport.js
martenrichter: good..

src/webtransport.js
martenrichter: ok, but it should never be null, only if there is an error in the code. but ok

src/webtransport.js
achingbrain: Do you mean this?:

```js
const loop = Http3EventLoop.createGlobalEventLoop()
```

`createGlobalEventLoop` only creates the loop if it doesn't exist so it's safe to call even if there's a loop already.

That said this is easier to read:

```js
const loop = Http3EventLoop.globalLoop ?? Http3EventLoop.createGlobalEventLoop()
```

..so I've changed it back to this.

As a future improvement maybe `createGlobalEventLoop` should throw if it's called twice?

src/types.ts
martenrichter: ok

src/webtransport.js
martenrichter: Mine is secondary also client. (Audio video router, browser as well as a network of multiple node.js).

And yes this should cover everything once it is done.
Now, back to your arm mac  problem

src/webtransport.js
martenrichter: > As a future improvement maybe `createGlobalEventLoop` should throw if the loop already exists? My opinion on that is not that strong.

No, this would not be an improvement. Currently, the event loop works as follows.
If there is no event, one is created, but additionally, a guarding is started, which checks, if it is not needed anymore for some time (a couple of 10 seconds) and then it destructs itself. In this way, there is only an event loop if it is needed.

src/webtransport.js
achingbrain: I've done the commenting and added a block at the bottom of the readme

martenrichter: Perfect, I will now turn on the unit tests..., and see if all platforms are working. You might want to check out the branch about OSX compilation. I try to do the universal binary in cmakelists.txt directly, so far no success, but I do not have a mac and rely on github runner and this takes some time.

martenrichter (approved):

PR #96, opened 2022-10-29
Try to create arm prebuilds

martenrichter:


achingbrain:
Well this is tedious - https://boringssl.googlesource.com/boringssl/+/546093ca4a6b2bf9b7c29018687d53f8a2ad428a/CMakeLists.txt#300

In the olden times (e.g. during the ppc -> intel switch) you'd use a tool called [lipo](https://ss64.com/osx/lipo.html) to make a universal binary from a ppc and an intel build.  If boringssl only supports building one arch at a time that might be necessary?

martenrichter:
> Well this is tedious - https://boringssl.googlesource.com/boringssl/+/546093ca4a6b2bf9b7c29018687d53f8a2ad428a/CMakeLists.txt#300
> 
> In the olden times (e.g. during the ppc -> intel switch) you'd use a tool called [lipo](https://ss64.com/osx/lipo.html) to make a universal binary from a ppc and an intel build. If boringssl only supports building one arch at a time that might be necessary.

Well, I have a problem, the lines did not turn cmake into multi-arch.  So if boring does not support it anyway, I should cross-compile..., ok, I try again..... (sadly I did not find if one can access the arm GitHub runner beta test)

achingbrain:
Ok so `lipo` is still bundled with XCode.

Certainly if I build this module locally I can create a universal binary from the arm64 version `npm build` creates and the `x64` version downloaded from the latest release (I renamed the built `webtransport.node` file to `webtransport_arm.node` to make it easier for me):

```console
$ lipo -create webtransport_arm.node webtransport_x64.node -output webtransport.node
$ file webtransport.node 
webtransport.node: Mach-O universal binary with 2 architectures: [x86_64:Mach-O 64-bit dynamically linked shared library x86_64Mach-O 64-bit dynamically linked shared library x86_64] [arm64]
webtransport.node (for architecture x86_64):    Mach-O 64-bit dynamically linked shared library x86_64
webtransport.node (for architecture arm64):     Mach-O 64-bit dynamically linked shared library arm64
```

The universal binary works for running the tests in this module (previously needing an `arm64` build for me locally) and also in my libp2p test harness which for tedious reasons runs through rosetta so needs a `x86_64` build.

martenrichter:
> Ok so `lipo` is still bundled with XCode.
> 
> Certainly if I build this module locally I can create a universal binary from the arm64 version `npm build` creates and the `x64` version downloaded from the latest release (I renamed the built `webtransport.node` file to `webtransport_arm.node` and the downloaded version to `webtransport_x64.node` to make it easier for me):
> 
> ```
> $ lipo -create webtransport_arm.node webtransport_x64.node -output webtransport.node
> $ file webtransport.node 
> webtransport.node: Mach-O universal binary with 2 architectures: [x86_64:Mach-O 64-bit dynamically linked shared library x86_64Mach-O 64-bit dynamically linked shared library x86_64] [arm64]
> webtransport.node (for architecture x86_64):    Mach-O 64-bit dynamically linked shared library x86_64
> webtransport.node (for architecture arm64):     Mach-O 64-bit dynamically linked shared library arm64
> ```
> 
> The universal binary works for running the tests in this module (previously needing an `arm64` build for me locally) and also in my libp2p test harness which for tedious reasons runs through rosetta so needs a `x86_64` build.

I try now if I can cross-compile, but than we will have separate binaries.

martenrichter:
Should work now...

PR #97, opened 2022-10-29
fix: limit exposed types

achingbrain:
* Moves all .js to the `lib` dir
* Splits all classes out into invdividual files
* Only exports spec and `WebTransportSession` types, `Http3Server` and `WebTransport` classes, and the `testcheck` function

A small refactor was necessary - because the `WebTransport` class cannot be declared in the same context as the `WebTransport` interface, it needs to only expose the same fields as the interface.

This meant merging the `establishSession` session into it's construtor (good news: any errors thrown by it's floating promise will now be caught) and using a [WeakMap](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakMap) to store the `Http3WTSession` instance previously stored on `this.sessionint`.


achingbrain:
As promised: https://github.com/fails-components/webtransport/pull/94#discussion_r1008702476

martenrichter:
Please rebase with current master

achingbrain:
It's up to date already:

```console
% git log --format="%s %H"
fix: limit exposed types 7f8f72653a9de6ac888ab1fc10272520f3a0c509
Merge pull request #94 from achingbrain/feat/add-types 2be9bcab472c29bfc43e2f8c92efebc2543f2728
Merge branch 'master' into feat/add-types e96615cba742d30efa23e61e1f15f5c81dbb9fd3
chore: comment out unimplemented spec fields 11b7c2b62d7a3e8fa5a981db963cc98b22ea0454
chore: make event loop creation easier to read cc312983e55704962f17424ccd5689ce16a205c0
chore: remove extra logging 0bd693870d4849061c595127d69a7330e6ad4e67
```

achingbrain:
I think the build might be failing because of this? https://github.com/fails-components/webtransport/commit/7985f4837eb7ec3144c0ba556e0965480e04464f

martenrichter:
Yes the weekly libquiche update. But before the commit it builds. Also, did you make the split so that the blame history is preserved so that for example the pending Byob branches do not break?

achingbrain:
Interesting.  The error seems to be:

```
/home/runner/work/webtransport/webtransport/./src/http3serversession.h:110:37: error: ‘kRfcAndDraft04’ is not a member of ‘quic::HttpDatagramSupport’
  110 |         return HttpDatagramSupport::kRfcAndDraft04;
      |                                     ^~~~~~~~~~~~~~
make[2]: *** [CMakeFiles/webtransport.dir/build.make:132: CMakeFiles/webtransport.dir/src/http3dispatcher.cc.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:998: CMakeFiles/webtransport.dir/all] Error 2
```

Which isn't related to the changes here?

> Also, did you make the split so that the blame history is preserved so that for example the pending Byob branches do not break?

Probably not?  How would you do that?

martenrichter:
For example:
https://stackoverflow.com/questions/3887736/keep-git-history-when-splitting-a-file
but there may be better instructions. A bit tedious, but better to understand why something was changed. But anyway you can also just keep al internal stuff in one file and rename this.

Regarding the error you submitted the state of your submodules, see the diff, and this is outdated. So you missed a `git submodule update` before committing.

achingbrain:
Ok, build is happy now and I think I've preserved the blame info on all the new files.

martenrichter:
Looks good, but I am tired and will look tomorrow on it.

martenrichter:
> This meant merging the `establishSession` session into its constructor (good news: any errors thrown by it's floating promise will now be caught) and using a [WeakMap](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakMap) to store the `Http3WTSession` instance previously stored on `this.sessionint`.

I do not understand, why you are using a WeakMap to store the reference.
First, if we have only one reference, we would use a WeakRef, because no lookup time in the map.
Second, the reference was there for a reason, it was managing the lifecycle of the session. Now the session can go away anytime, even if the WebTransport object is still there, all randomly handled by the garbage collector.

So I think it has to be converted back. Or is there a problem with the nonweakref, that I do not see.

martenrichter:
The rest looks perfect, so that the last point is the only one.

achingbrain:
> Now the session can go away anytime, even if the WebTransport object is still there, all randomly handled by the garbage collector.

No, the value will only get evicted from the map if the key is collected.  The trick to `WeakMap` is that it doesn't create a strong reference to it's key, so the key can be garbage collected - when that happens the value will be evicted and not before.

> Edit: also close etc will crash, if the object is garbage collected

`close` etc will not crash, because the `Http3WTSession` value is only evicted from the map when the `WebTransport` key object itself is garbage collected, at which point there's no more `close` etc either.

`WeakMap` is how TypeScript supports [private class features](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields) for pre-ES2022 targets:

```ts
class Foo {
  #privateField: string
  
  constructor (value: string) {
      this.#privateField = value
      console.info(this.#privateField.toString())
  }
}
```

compiles to:

```js
var __classPrivateFieldSet = (this && this.__classPrivateFieldSet) || function (receiver, state, value, kind, f) {
    if (kind === "m") throw new TypeError("Private method is not writable");
    if (kind === "a" && !f) throw new TypeError("Private accessor was defined without a setter");
    if (typeof state === "function" ? receiver !== state || !f : !state.has(receiver)) throw new TypeError("Cannot write private member to an object whose class did not declare it");
    return (kind === "a" ? f.call(receiver, value) : f ? f.value = value : state.set(receiver, value)), value;
};
var __classPrivateFieldGet = (this && this.__classPrivateFieldGet) || function (receiver, state, kind, f) {
    if (kind === "a" && !f) throw new TypeError("Private accessor was defined without a getter");
    if (typeof state === "function" ? receiver !== state || !f : !state.has(receiver)) throw new TypeError("Cannot read private member from an object whose class did not declare it");
    return kind === "m" ? f : kind === "a" ? f.call(receiver) : f ? f.value : state.get(receiver);
};
var _Foo_privateField;
export class Foo {
    constructor(value) {
        _Foo_privateField.set(this, void 0);
        __classPrivateFieldSet(this, _Foo_privateField, value, "f");
        console.info(__classPrivateFieldGet(this, _Foo_privateField, "f").toString());
    }
}
_Foo_privateField = new WeakMap();
```

Note that when returning the value from `state.get(receiver)`, `__classPrivateFieldGet` doesn't guard on it being undefined.

> Or is there a problem with the nonweakref, that I do not see?

It adds a `sessionint` field to the WebTransport class.  If a function expects an instance of the `WebTransport` class you then cannot pass an object that conforms to the `WebTransport` spec interface since the spec interface doesn't have the `sessionint` field.

martenrichter:
Ok, I thought it was the other way around with WeakMap, then it is ok. I should have looked it up.
And you can not have hidden fields in an interface? Not also not by using a derived class? (maybe I am too much a c++ developer with polymorphism in mind, who still stumbling through js and with no idea about ts).
Anyway, then it is fine.

PR #98, opened 2022-10-29
test: add test for connecting to server that is not running

achingbrain:
Not sure if I'm doing this right, but trying to connect to a server that isn't running never errors, it just hangs `await`ing the `WebTransport` ready promise forever.


test/test.js
```diff
@@ -18,6 +18,20 @@ async function run() {
...
+  let badClient = new WebTransport('https://127.0.0.1:49823/echo', {
+    serverCertificateHashes: [{ algorithm: 'sha-256', value: Buffer.from('a589bf4f98a0158aa890328d5d3f519b9e2a5b1e61b09eb10b7a9be0e79bf148', 'hex') }]
+  })
+  await badClient.ready
+    .then(() => {
+      console.error('Successfully connected to a non-running server?!')
+      process.exit(1)
+    })
+    .catch(() => {
+      console.log('Did not connect to non-running server')
```
achingbrain, line 32: Would be nice to assert on the thrown error here - `.code` property or `.message` but no error is throw?

martenrichter: There is currently no mechanism in `Http3Client`.
Probably a time out in `handleConnecting` together with code to tear down the connection in an error case.
But agreed, should be done. However currently I want first to write the code, that needs the time out.

martenrichter: But you can program a timeout yourself on the js side and close after time out.
Something like setTimeout(()=>{if (notconnected) session.close()},10000), so I do not see urgency.

martenrichter: After being more awake. I think the main problem is that no one is responding. Since the underlying UDP protocol is connectionless, the only way to detect that no one is responding is via timeout in the handshakes in the client code.
The question is what should be the default timeout?

achingbrain: From experimentation it's 4s in Chrome.  I can't see anything in the spec about overriding default timeouts though, which raises an eyebrow slightly.

martenrichter: Since it is not in the specs, we would deviate, if we allow it to change.
We can implement it easily on the c++ side, (but my development dir is currently tied to the macOS arm build...., so not now), but then it is fixed, or we allow to pass a different value at construction time extending the spec.

martenrichter:
Merge this patch is currently impossible since the test fails and the tests protect the master branch.

martenrichter:
Step to towards passing this test https://github.com/fails-components/webtransport/pull/101

martenrichter:
The time out fix seems to have errors

martenrichter:
This https://github.com/fails-components/webtransport/pull/102 should allow integration of the test. Tests are running.

achingbrain:
Probably best to add this test suite to the new version in #103 - I'll open a new PR with this test in it once that is merged.

martenrichter:
> Probably best to add this test suite to the new version in #103 - I'll open a new PR with this test in it once that is merged.

Sure but first it was to work...., so merging the test suite will also wait until a problem in the underlying code is fixed.

martenrichter:
I will merge it anyway. It showed an issue with the event loop that is now resolved. So a good reference test for the new tests.

martenrichter:
OK, units tests say Linux is ok. But Mac OS and Windows fail... I will add a timeout for these platforms.

PR #102, opened 2022-10-30
Fix clienttime, error throwing and linter happy, also revert some changes after types

martenrichter:


martenrichter:
Nice there is a seg fault in CI/CD but not in my local machine (Release vs. Debug)

martenrichter:
Segfault still there, but only in release builds....

martenrichter:
TODO split delegate from client

lib/client.js
```diff
@@ -94,7 +93,7 @@ export class Http3Client extends Http3WebTransport {
   /**
    * @param {ClientConnectedEvent} args
    */
-  onClientConnected (args) {
```
achingbrain, line 97: Is it worth adding a linter for this?  [standard](https://www.npmjs.com/package/standard) or something?

lib/session.js
```diff
@@ -395,6 +399,7 @@ export class Http3WTSession {
           visitor.onClose(args)
           break
         case 'DatagramReceived':
+          // eslint-disable-next-line no-prototype-builtins
           if (visitor && args.hasOwnProperty('datagram'))
```
achingbrain, line 403: This should pacify the linter:

```suggestion
          if (visitor && Object.prototype.hasOwnProperty.call(args, 'datagram'))
```

lib/session.js
```diff
@@ -404,7 +409,9 @@ export class Http3WTSession {
         case 'Http3WTStreamVisitor':
           if (
             visitor &&
+            // eslint-disable-next-line no-prototype-builtins
             args.hasOwnProperty('bidirectional') &&
```
achingbrain, line 413:
```suggestion
            Object.prototype.hasOwnProperty.call(args, 'bidirectional') &&
```

lib/session.js
```diff
@@ -404,7 +409,9 @@ export class Http3WTSession {
...
+            // eslint-disable-next-line no-prototype-builtins
             args.hasOwnProperty('incoming')
```
achingbrain, line 415:
```suggestion
            Object.prototype.hasOwnProperty.call(args, 'incoming')
```

lib/session.js
achingbrain, line 403: This should pacify the linter:

```suggestion
          if (visitor && Object.prototype.hasOwnProperty.call(args, 'datagram'))
```

lib/stream.js
```diff
@@ -295,6 +299,7 @@ export class Http3WTStream {
           visitor.onStreamRecvSignal(args)
           break
         case 'StreamRead':
+          // eslint-disable-next-line no-prototype-builtins
           if (visitor && args.hasOwnProperty('data')) {
```
achingbrain, line 303:
```suggestion
          // eslint-disable-next-line no-prototype-builtins
          if (visitor && Object.prototype.hasOwnProperty.call(args, 'data')) {
```

lib/client.js
martenrichter: There should be already some linter config files in the project. So far I do not protect the branch with the linter, but I always check against it manually before committing.

lib/session.js
martenrichter: Sure this will help

lib/stream.js
martenrichter: Prefect

martenrichter:
Should work now, after this the test PR has to be rebased.

lib/client.js
achingbrain: Might be worth adding an npm script for it and running it in CI.  Will submit a PR.

martenrichter: Ok, good idea, had just not a high priority, while I was alone.

achingbrain: Completely fair!

PR #103, opened 2022-10-31
feat: add test framework, address method and server ready/closed promises

### LINKED ISSUE #100: Testing
The test suite is one function that reuses the same server and runs the tests sequentially.  This can be improved - would you accept a PR that uses a test framework, splits the tests into separate files so we can run them individually and with set-up/tear-down steps for better isolation, etc?

My(weak) preference is for [mocha](https://mochajs.org/), but I see there's now a [built-in test framework](https://nodejs.org/api/test.html) in the most recent node, which might be fun to experiment with, but it will restrict development to node v18.0+, which might not be a big deal in the long run, as long as the built binaries still run on older versions 🤷.

achingbrain:
* Adds mocha test framework
* Adds `.address()` method to `Http3Server` to know which port/host it ends up listening on
* Adds `.ready` and `closed` promises to `Http3Server` so you don't have to guess when it's ready to use

The `.address()` method is necessary to know which port a server listens on when you specify `0` as the parameter to avoid clashes during testing.

The `.ready` and `.closed` promises are triggered by callbacks from the C++ code - I hope I implemented these correctly.

I've replicated all of the tests from the current suite, these can be deleted in a follow up PR.

Because we don't have to use an arbitrary timeout to wait for the server to be ready the test suite is now much faster to run - on my machine the port is opened in 20-30ms so no need to wait 2000ms as in the current suite.

The event loop still takes a long time to shut down though, maybe this can be refactored?

Fixes #100


lib/webtransport.js
```diff
@@ -1,6 +1,7 @@
 import { Http3Client } from './client.js'
 import { Http3WTSession } from './session.js'
 
+/** @type {WeakMap<WebTransport, Http3WTSession>} */
```
achingbrain, line 4: I forgot this in the previous PR, sorry!

lib/webtransport.js
```diff
@@ -69,14 +70,35 @@ const sessions = new WeakMap()
    * @param {WebTransportCloseInfo} [closeinfo]
    */
   close(closeinfo) {
-    return sessions.get(this).close(closeinfo)
+    const session = sessions.get(this)
```
achingbrain, line 73: Now tsc knows what type the `WeakMap` is, we need to null-guard on the values, even though they'll be defined.

lib/server.js
```diff
@@ -20,6 +30,12 @@ export class Http3Server extends Http3WebTransport {
 
     /** @type {Record<string, ReadableStreamController<any>>} */
     this.sessionController = {}
+
+    this._ready = defer()
```
achingbrain, line 34: This adds a `_ready` field to the server.  We may wish to store these in a `WeakMap` instead to hide them from the types or better (I think - more familiar at least) switch to an `EventEmitter` model for `open`/`connect`/`close`/`end`/`error` events for the server side like the http2 server in node core.

martenrichter:
Thanks, it looks promising, I will try to comment on it shortly.

martenrichter:
> Because we don't have to use an arbitrary timeout to wait for the server to be ready the test suite is now much faster to run - on my machine the port is opened in 20-30ms so no need to wait 2000ms as in the current suite.

Well, this is true. But what happens, if the server or client crashes very fast? So it is also a test if one waits some time to spot this.
Btw, there were some issues by davedoesdev, where it may be a good idea to add a similar test code to the tests.

lib/server.js
martenrichter: Well, I would like to stick with the more modern approach. Since it is similar to the sessions etc in WebTransport. But you have a point many developers may be more familiar with the old stuff and have code compatible with this. If you like you can always make a wrapper package that maps the server, client, and sessions for this package and also the browser to an old-fashioned interface. There are probably some people who like it. But I actually like the promise-based steam-style interface of webtransport, after first hating it.

lib/server.js
```diff
@@ -35,6 +51,22 @@ export class Http3Server extends Http3WebTransport {
     this.stopped = true
   }
 
+  /**
+   * @returns {{ port: number, host: string, family: 'IPv4' | 'IPv6' } | null}
```
martenrichter, line 55: This implementation works but is very resource intensive. Every call from js to C++ costs a lot of time. So it would be better to keep a copy of the address and port and host on the js side and update it, e.g. when the server goes up.

lib/server.js
```diff
@@ -53,29 +85,59 @@ export class Http3Server extends Http3WebTransport {
...
+      const sesobj = new Http3WTSession({
+        object: args.session,
+        parentobj: this
+      })
+      if (this.sessionController[args.path])
+        this.sessionController[args.path].enqueue(sesobj)
+    } else throw new Error('Http3WTSessionVisitor')
+  }
+
+  /**
```
martenrichter, line 102: Please merge the following functions into one and use only one event.
Furthermore, it should include an object that may also contain information about the change (i.e. host and address).

lib/server.js
```diff
@@ -53,29 +85,59 @@ export class Http3Server extends Http3WebTransport {
...
-          if (args.object) {
-            const sesobj = new Http3WTSession({
-              object: args.session,
-              parentobj: this
-            })
-            if (this.sessionController[args.path])
-              this.sessionController[args.path].enqueue(sesobj)
-          } else throw new Error('Http3WTSessionVisitor')
-
+          this.onHttp3WTSessionVisitor(args)
```
martenrichter, line 131: Same as above

lib/types.ts
```diff
@@ -126,6 +126,48 @@ export interface Http3ClientEventHandler {
...
 
+export interface Http3WTServerSessionVisitorEvent extends Http3WTSessionVisitorEvent {
+  path: string
+  object: any
+}
+
+/**
+ * The Http3 server is listening on the specified port
+ */
+export interface Http3ServerListeningEvent {
```
martenrichter, line 137: Again please merge

src/http3eventloop.cc
```diff
@@ -295,6 +295,33 @@ namespace quic
       progress_->Send(&report, 1);
   }
 
+  void Http3EventLoop::informHttp3ServerListening(Http3Server *serverobj)
```
martenrichter, line 298: Two things, please move it up where the other Http3Server callbacks are, they are actually sorted a bit...
Second, merge the three functions into `informHttp3ServerStatus` with an enum indicating the purpose and maybe some data. There is actually also a mechanism that allows passing memory that is later deleted on the other thread. This way also extended status information can be added to the object passed to the js side. (Also changes in napi may require editing all handlers, so I try to keep them at a minimum....)

src/http3eventloop.cc
```diff
@@ -631,6 +658,48 @@ namespace quic
     cbsession_.Call({retObj});
   }
 
+  void Http3EventLoop::processHttp3ServerListening(Http3Server *serverobj)
+  {
+    if (!checkQw()) return;
```
martenrichter, line 663: Again please merge,

src/http3eventloop.cc
```diff
@@ -688,6 +757,21 @@ namespace quic
         processStream(false, false, cur.sessionobj, cur.stream);
       }
       break;
+      case Http3ProgressReport::ServerError:
```
martenrichter, line 760: Please merge

martenrichter: See cur.para for the mechanism of passing data.

src/http3eventloop.h
```diff
@@ -69,6 +69,9 @@ namespace quic
             IncomUniDiStream,
             OutgoBiDiStream,
             OutgoUniDiStream,
+            ServerError,
```
martenrichter, line 72: merge...

src/http3server.cc
```diff
@@ -398,4 +402,17 @@ namespace quic
     }
   }
 
+  Napi::Value Http3ServerJS::port(const Napi::CallbackInfo &info)
```
martenrichter, line 405: Please move the functionality to the server status update. Calls from js into this object are rather slow.

test/unidirectional-streams.spec.js
```diff
@@ -0,0 +1,114 @@
+import { createServer } from './fixtures/server.js'
```
martenrichter, line 1: Looks good, consider adding some test cases from resolved seg fault issues.

martenrichter:
> The event loop still takes a long time to shut down though, maybe this can be refactored?

This is intentional, since tearing it up and down also costs CPU time and memory, so it waits for a reasonable time like a hard drive powersave mode, which requires some idle time.
And I have a compilation time of around 20-50 min in CI/CD. So I do not care about the time the other tests take...

martenrichter:
What I am missing, is a test if the eventloop is gone. It is actually a crucial test, that we do not have a memory leak and that the objects are destructed. Very critical.

lib/server.js
martenrichter: And there is another thing to consider. The objects have members, that should only be access from the libquiche eventloop or the js side on object creation, but never from js side after creation. This applies to any member, that is potentially altered from the libquiche eventlooop. The reason is thread safety without a locking mechanism. And this way of accessing the members break this principle, which may or may not lead to random segfaults.

test/fixtures/read-stream.js
```diff
@@ -0,0 +1,37 @@
+
+/**
```
martenrichter, line 2: This applies to all tests. Is there a mechanism, that will fail the test after a certain time, if it is hanging. I do not spot one at first glimpse. But this is crucial, in order to exit the CI/CD pipeline early and not waste resources and also to give the developer immediate feedback. The old tests have this mechanism. (And such problems occur very often during development on some platforms...). Also a check of the runtime, that it does not exceed a certain value would be good, to spot performance problems with the two eventloops.

achingbrain: Mocha has built in timeouts for all tests, the default is 2s.

It can be overridden per suite with:

```js
describe('a module', function () {
  this.timeout(5000) // all tests in the suite will now timeout after 5s

  it('should do a thing', () => {
    // ...
  })
})
```

Or per test with:

```js
describe('a module', () => {
  
  it('should do a thing', function () {
    this.timeout(5000) // only this test will timeout after 5s, all others use the default of 2s
    // ...
  })
})
```

martenrichter: Perfect, because these timeouts are really important for the test. It is really often, that something blocks.

lib/server.js
achingbrain: Calling into C++ to get the server address is [what node core does](https://github.com/nodejs/node/blob/main/lib/net.js#L1680-L1692) in the same call?  I take the point that it's slower than caching the values on the js side but this isn't the kind of method you invoke very often.

A interesting point about thread safety, but once the server is running the host and port values aren't going to change?

martenrichter: Well, I do not know, for the server probably not. 
But maybe you want the same for the client or other attributes from the server later and these can change,
since quic allows to change IP etc during the connection. So it is better to build in a way, that it can be extended in the future and the philosophy on all part is similar in coding and calling conventions.

martenrichter: Another point, we do not know, is if the host and port are always set within the js time. At a later point, this may move into the other thread, and if you call it before the server is up...

martenrichter: Seriously, those last seg fault issues were all of the kind, and yes for similar stuff (maybe not this one), tests may help.
And some took days and hours to find.
That is why I am a bit strict in the design pattern of these objects.

lib/server.js
martenrichter: Sure I have looked at it, but it progresses slowly and it is hard to guess, what and when it comes.
But I also do not know, if the initial http/3 support has webtransport support.
So that is nothing I worry about currently.

martenrichter:
> > What I am missing, is a test if the eventloop is gone.
> 
> I didn't see one in the existing tests so that's why there's not one here - I've added a basic one, please let me know if it's not sufficient.

Well, there is one, but I think I was not describing it well enough to you.
This:
` setTimeout(() => {
    if (!testcheck()) {
      console.log('tests took too long, probably hanging')
      process.exit(1)
    } else {
      console.log('global event loop gone, everything alright')
      process.exit(0)
    }
  }, 40 * 1000)`
is actually a test.
So all we need is a test, which runs as the last test and which waits long enough, that the event loop can go away.
In this way, we test all objects, servers, clients, and sessions if they properly deregister from the event loop and if their js side is dereferencing hopefully everything. (A check that the deletion of the c++ object does not crash, was not possible, since the GC is not deterministic.) This actually important for detecting potential memory leaks.

lib/server.js
martenrichter: Also, I assume that the http/3 interface would be different from http/2 (as HTTP/2 differs from HTTP), just because the protocol itself is different.
Just out of curiosity, do you want only to use WebTransport or also http/3 (not implemented), I assume only webtransport?

Anyway, I believe, that can one make a wrapper that unpromisifies the promises and streams into callbacks.
Btw. here is an article describing the advantages of promises:
https://dev.to/somedood/promises-and-events-some-pitfalls-and-workarounds-elp
probably these things are the reason the whole webtransport API is designed as it is.

package.json
```diff
@@ -11,7 +11,7 @@
   "types": "./dist/lib/index.d.ts",
   "scripts": {
     "start": "node test/echoserver.js",
-    "test": "node test/test.js",
```
martenrichter, line 14: please keep the oldtest as an `oldtest` script.
I will invoke them after the new tests in CI/CD and if yours pass and the old fails, we see, what needs to be improved. So testing the tests.

martenrichter:
Btw. would you midn if the test are also incorporated into https://github.com/fails-components/webtransport-ponyfill-websocket .
It is the sister project, sharing part of the code base (so I also have to bring to sync with this project).
It provides a fallback over websocket, if webtransport is not available, providing the almost same interface and also no head of line blocking. Btw. webtransport over http/2 I found the spec, actually I was wrong it is not a legacy thing but a future spec, but so far so sign in libquiche, that it is implemented.

martenrichter:
Interesting 4 tests are failing on a debug build...., will investigate.

test/bidirectional-streams.spec.js
```diff
@@ -0,0 +1,118 @@
...
+import { defer } from '../lib/utils.js'
+
+/**
+ * @template T
+ * @typedef {import('../lib/types').Deferred<T>} Deferred<T>
+ */
+
+const SERVER_PATH = '/unidirectional-streams'
+
+describe('bidirectional streams', function () {
```
martenrichter, line 16: Linter complains here and on other lines, that describe is not defined...., I am also puzzled from where the definition should come. One can disable it globally for the file, but normally it is very handy

martenrichter:
Bidirectional stream fails, since stopServer takes forever in Linux Docker container on Windows...
Datagram is another issue

martenrichter:
Bidi stream can be solved with a larger timeout, but I have the feeling the streams are not closed properly.....
The datagrams just failed because the test failed before...., propabaly a port was still open...

martenrichter (approved):
OK, I have addresses most issues in a PR , which I will commit shortly after merging this one.
Open points: 

- missing eventloop check, (so if all objects tear down correctly
- can the test also go to the sister project
- and there seems to be an issue below one bidirectional test, I think it is a missing cancel on the stream, it takes a long time to shutdown.

test/bidirectional-streams.spec.js
achingbrain: Mocha injects these into the global scope, with eslint you can put a comment at the top of the file to tell it the file is a mocha test:

```js
/* eslint-env mocha */
```

achingbrain:
> This:
`setTimeout(() => { if (!testcheck()) { console.log('tests took too long, probably hanging') process.exit(1) } else { console.log('global event loop gone, everything alright') process.exit(0) } }, 40 * 1000)`
is actually a test.

Ah, right - I saw that but thought it was redundant considering mocha has built-in timeouts for all the tests.  I think the added `event-loop.spec.js` covers it?  It asserts that you can create an event loop, and that when you shut the event loop down, the `globalLoop` reference goes away which is what the `testcheck` function does.

achingbrain:
> I think the added event-loop.spec.js covers it?

I say that, but it looks like you've turned that test off in [another commit](https://github.com/fails-components/webtransport/commit/bc1734bce140374a3403681b9cbfea4e2431cc49)?

If you don't want something to run, you can just do:

```js
describe.skip('event-loop', () => {
```

Though I'm curious as to why it was necessary to turn the tests off entirely?

achingbrain:
> > The event loop still takes a long time to shut down though, maybe this can be refactored?
>
>This is intentional, since tearing it up and down also costs CPU time and memory, so it waits for a reasonable time like a hard drive powersave mode, which requires some idle time.
> And I have a compilation time of around 20-50 min in CI/CD. So I do not care about the time the other tests take...

I think it depends on your style of development. I run tests a lot so the extra wait time mounts up very quickly.

The other thing is that it adds shutdown time to applications using this module which tends to raise eyebrows.  When webtransport/http3 finally makes it into node core, there will almost certainly not also be a seemingly arbitrary 10 second shutdown delay added alongside it, so there must be a better way.

martenrichter:
> > I think the added event-loop.spec.js covers it?
> 
> I say that, but it looks like you've turned that test off in [another commit](https://github.com/fails-components/webtransport/commit/bc1734bce140374a3403681b9cbfea4e2431cc49)?
> 
> If you don't want something to run, you can just do:
> 
> ```js
> describe.skip('event-loop', () => {
> ```
> 
> Though I'm curious as to why it was necessary to turn the tests off entirely?

On windows builds mocha returned an exit code 1, if any of the task, that shutdown the event loop were present. Though all test passed. I believe it is a seg fault or something else. Forcefully shuting down the event loop is also not support. Since I assumed during coding, that the eventloop is present until all objects, that depend on it are garbage collected. So it is only allowed to shut it off, if all objects, that depend on it are gc'ed. At least, that is, what I believed had happened. That the other platforms passed, is just random. So the test for the eventloop is in away a test, if the lifecycle management does not fail.

martenrichter:
> > > The event loop still takes a long time to shut down though, maybe this can be refactored?
> > 
> > 
> > This is intentional, since tearing it up and down also costs CPU time and memory, so it waits for a reasonable time like a hard drive powersave mode, which requires some idle time.
> > And I have a compilation time of around 20-50 min in CI/CD. So I do not care about the time the other tests take...
> 
> I think it depends on your style of development. I run tests a lot so the extra wait time mounts up very quickly.
> 
> The other thing is that it adds shutdown time to applications using this module which tends to raise eyebrows. When webtransport/http3 finally makes it into node core, there will almost certainly not also be a seemingly arbitrary 10 second shutdown delay added alongside it, so there must be a better way.

Well the setInterval for checking the state can be shorter. But the main problem is lifetime management of the C++ objects through the GC (it adds to the delay). Of course the ties to the eventloop can be broken earlier, but then carefully checks in the C++ objects have to be added and I think it will be though to not miss anything and to provoke a seg fault sometime and not spotable by the tests.
(One main problem of the implementation is that it runs a separate event loop, on the other hand it is also an advantage, since the low level protocol work runs off the main thread.....

PR #110, opened 2022-11-05
chore: add linter script and run it in CI

achingbrain:
Follow up from https://github.com/fails-components/webtransport/pull/102#discussion_r1010444739

* Adds a `"lint"` npm script that runs `eslint` against the `lib` and `test` directories.
* Adds dirty chai to enable `expect(foo).to.be.true()`-style assertions as eslint objects to `expect(foo).to.be.true`
* Runs linting in CI on linux
* Fixes all linting issues
* Skips the event-loop tests instead of renaming the file
* Removes the `after` that stops the event loop instead of renaming the file
* Adds the `eslint-env moch` directive to each test file instead of turning off the `no-undefined` rule

### DIFF
--- a/.github/workflows/libtest-linux.yml
+++ b/.github/workflows/libtest-linux.yml
@@ -25,19 +25,22 @@ jobs:
       # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
       - uses: actions/checkout@v3
         with:
-          submodules: recursive    
-        
+          submodules: recursive
+
       - name: Set up NodeJS
         uses: actions/setup-node@v1
         with:
           node-version: 16
-        
+
       - name: Install dependencies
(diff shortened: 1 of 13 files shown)


martenrichter:
Looks good. Merge will take some time, the number of runners of CI/CD is limited.....

.github/workflows/libtest-linux.yml
```diff
@@ -25,19 +25,22 @@ jobs:
...
           node-version: 16
-        
+
       - name: Install dependencies
         run: npm install
 
       - name: Build library
         run: npm run build
-        
+
```
martenrichter, line 40: I will merge now, but probably move the linter up. Since the build takes 20-30 minutes I do not want to wait for it to fail.

PR #111, opened 2022-11-05
fix: export http3 server init type

achingbrain:
Adds a type for the args you can pass to the http3server constructor and exports it from the module

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -22,6 +22,7 @@ import { Http3EventLoop } from './event-loop.js'
  *
  * Public API
  * @typedef {import('./types').WebTransportSession} WebTransportSession
+ * @typedef {import('./types').Http3ServerInit} Http3ServerInit
  */
 
 export function testcheck() {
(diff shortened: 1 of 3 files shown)


martenrichter:
Looks good, auto-merge is enabled, but will take some time, the MacOS runners are busy...

martenrichter:
Please do not merge the master to other PR. I will do that. I am running out of mac os x runners, which I need for testing the binary build CI for arm. I think I have only 4, and every check needs 2 at the moment, and I like to solve the prebuild stuff this weekend.

martenrichter:
> Please do not merge the master to other PR. I will do that. I am running out of mac os x runners, which I need for testing the binary build CI for arm. I think I have only 4, and every check needs 2 at the moment, and I like to solve the prebuild stuff this weekend.

Since the binary pipeline is fixed, the above does not apply anymore. (Was just because a release build needs many runners...)

PR #112, opened 2022-11-05
chore: add scripts to build mac universal binary

### LINKED ISSUE #105: Finding icu4c during build on M1 Macs
When I build master on my M1 Mac it fails to find `icu4c` with:

```
CMake Error at /usr/local/Cellar/cmake/3.24.2/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
  Failed to find all ICU components (missing: ICU_LIBRARY) (found version
  "70.1")
Call Stack (most recent call first):
  /usr/local/Cellar/cmake/3.24.2/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:594 (_FPHSA_FAILURE_MESSAGE)
  /usr/local/Cellar/cmake/3.24.2/share/cmake/Modules/FindICU.cmake:335 (FIND_PACKAGE_HANDLE_STANDARD_ARGS)
  CMakeLists.txt:38 (find_package)
```

It starts to work if I change [this line](https://github.com/fails-components/webtransport/blob/master/CMakeLists.txt#L34) in `CMakeLists.txt` from:

```
                SET(ICU_ROOT "/opt/homebrew/icu4c")
```

to:

```
                SET(ICU_ROOT "/usr/local/opt/icu4c")
```

I don't know if this is down to my homebrew version or something else but as CI seems happy with the config I'm posting this here in case it helps someone else in the future.

```console
% brew --version
Homebrew 3.6.7
```

In that file the `ICU_ROOT` gets set to `/usr/local/opt/icu4c` if the architecture is `x86_64` which it's not in my case so maybe the config is expecting an older homebrew version?

achingbrain:
To aid local development, adds `build-universal` and `rebuild-universal` npm scripts to build universal binaries on Mac OS X.

Refactors the attempt to find compiled WebTransport binaries to search known locations in order to ensure downloaded production versions are found first, but if not, fall back to development versions including a universal version.

Adds tests for the binary resolution.

Fixes #105

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -31,7 +31,7 @@ IF( NOT WIN32)
                 SET(ICU_ROOT "/usr/local/opt/icu4c")
             ELSE()
                 MESSAGE(STATUS "set icu root homebrew")
-                SET(ICU_ROOT "/opt/homebrew/icu4c")
+                SET(ICU_ROOT "/usr/local/opt/icu4c")
             ENDIF()
         ENDIF()
     ENDIF(APPLE)
@@ -82,7 +82,7 @@ endif(COMMAND cmake_policy)
 
 
 
-#protobuf_generate_cpp(PROTO_SRCS PROTO_HDRS 
+#protobuf_generate_cpp(PROTO_SRCS PROTO_HDRS
 #third_party/quiche/quiche/quic/core/proto/cached_network_parameters.proto
 #third_party/quiche/quiche/quic/core/proto/crypto_server_config.proto
 #third_party/quiche/quiche/quic/core/proto/source_address_token.proto)
@@ -94,7 +94,7 @@ add_compile_definitions(QUICHE_ENABLE_LIBEVENT)
 
 IF (WIN32)
 
-add_library(icu STATIC 
+add_library(icu STATIC
 #icu
     third_party/icu/icu4c/source/common/appendable.cpp
     third_party/icu/icu4c/source/common/bytestream.cpp
@@ -171,7 +171,7 @@ add_library(icu STATIC
     ICU_DATA_DIR_WINDOWS=""
     )
 
-    target_compile_definitions(icu 
+    target_compile_definitions(icu
     PRIVATE U_DEFINE_FALSE_AND_TRUE=1
     U_STATIC_IMPLEMENTATION
     U_COMMON_IMPLEMENTATION
@@ -179,7 +179,7 @@ add_library(icu STATIC
 
 ENDIF(WIN32)
 
-add_library(gquiche STATIC 
+add_library(gquiche STATIC
 #plattform impl
 platform/quiche_platform_impl/quiche_export_impl.h
 platform/quiche_platform_impl/quiche_logging_impl.h
@@ -937,7 +937,7 @@ DEPENDS protobuf::protoc ${filename} ${CMAKE_CURRENT_SOURCE_DIR}/third_party/qui
 ${CMAKE_CURRENT_SOURCE_DIR}/third_party/quiche/quiche/quic/core/proto/crypto_server_config.proto
 ${CMAKE_CURRENT_SOURCE_DIR}/third_party/quiche/quiche/quic/core/proto/source_address_token.proto
 COMMAND protobuf::protoc --proto_path=${PROTOMODEL_PATH} --cpp_out=${PROTOBINDING_PATH} ${CMAKE_CURRENT_SOURCE_DIR}/third_party/quiche/quiche/quic/core/proto/cached_network_parameters.proto
-COMMAND protobuf::protoc --proto_path=${PROTOMODEL_PATH} --cpp_out=${PROTOBINDING_PATH} ${CMAKE_CURRENT_SOURCE_DIR}/third_party/quiche/quiche/quic/core/proto/crypto_server_config.proto 
+COMMAND protobuf::protoc --proto_path=${PROTOMODEL_PATH} --cpp_out=${PROTOBINDING_PATH} ${CMAKE_CURRENT_SOURCE_DIR}/third_party/quiche/quiche/quic/core/proto/crypto_server_config.proto
 COMMAND protobuf::protoc --proto_path=${PROTOMODEL_PATH} --cpp_out=${PROTOBINDING_PATH} ${CMAKE_CURRENT_SOURCE_DIR}/third_party/quiche/quiche/quic/core/proto/source_address_token.proto
 )
 
@@ -957,10 +957,10 @@ ${CMAKE_CURRENT_SOURCE_DIR}/third_party/quiche/quiche/quic/core/proto/source_add
 #  LANGUAGE
 #    cpp
 #  GENERATE_EXTENSIONS
-#    .pb.h 
+#    .pb.h
 #    .pb.cc
 #  OUT_VAR PROTO_SRCS
-#  IMPORT_DIRS 
+#  IMPORT_DIRS
 #    "${CMAKE_CURRENT_SOURCE_DIR}/third_party/quiche"
 #  PROTOC_OUT_DIR
 #    ${CMAKE_CURRENT_SOURCE_DIR}
@@ -995,20 +995,20 @@ ELSE(WIN32)
     target_link_libraries(gquiche ICU::uc ICU::i18n)
 ENDIF(WIN32)
 
-target_link_libraries(gquiche 
-absl::base 
+target_link_libraries(gquiche
+absl::base
 absl::int128
-absl::strings 
+absl::strings
 absl::str_format
 absl::flat_hash_map
-absl::any 
+absl::any
 absl::synchronization
 absl::time
 absl::status
 absl::statusor
 absl::cleanup
 protobuf::libprotobuf
-ssl 
+ssl
 zlibstatic
 crypto
 event_core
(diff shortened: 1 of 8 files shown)


martenrichter:
Well, it fails on all other platforms, including normal mac. The output of the CI does not give a clue.
Anyway, it is interfering with https://github.com/fails-components/webtransport/pull/109 (I now the name of the branch is misleading). If it works, I will have a release candidate with a prebuild arm mac. But first the prebuild has to take place, takes minimum about an hour.  Edit: of course it failed.

CMakeLists.txt
```diff
@@ -31,7 +31,7 @@ IF( NOT WIN32)
                 SET(ICU_ROOT "/usr/local/opt/icu4c")
             ELSE()
                 MESSAGE(STATUS "set icu root homebrew")
-                SET(ICU_ROOT "/opt/homebrew/icu4c")
+                SET(ICU_ROOT "/usr/local/opt/icu4c")
```
martenrichter, line 34: So we do not need the if else at all?
I am still wondering, how it fits to the homebrew docs. Can you check the binary type of the icu lib in this dir? So it it is an universal binary?

lib/utils.js
```diff
@@ -21,3 +26,41 @@ export function defer () {
...
+ *
+ * @param {boolean} [includeDebug]
+ * @returns
+ */
+export function findWTBinaryPaths (includeDebug) {
+  const dirname = url.fileURLToPath(new URL('..', import.meta.url))
+
+  /** @type {string[]} */
+  const wtpaths = [
+    `${dirname}build/Release/webtransport.node`, // precompiled version
```
martenrichter, line 42: I am not sure. but I would prefer that the local build version always wins. It looks like in this case the precompiled always wins.

achingbrain: That was the intention, to favour the happy path of users who have installed the module on a platform that has a precompiled version available, rather than developers who will not have the precompiled version present.

martenrichter: Ok, but you have to trigger a build manually. So the happy folks would never be bothered, only if they trigger a build. 
So maybe switching the lines and adding a warning, if the local build is used?

CMakeLists.txt
achingbrain: It looks like it's `x86_64`?

```
% file /usr/local/opt/icu4c/lib/libicudata.71.1.dylib 
/usr/local/opt/icu4c/lib/libicudata.71.1.dylib: Mach-O 64-bit dynamically linked shared library x86_64
```

```
% file /usr/local/opt/icu4c/bin/icuinfo 
/usr/local/opt/icu4c/bin/icuinfo: Mach-O 64-bit executable x86_64
```

martenrichter: Then it is probably bad for an arm build.
Also, icu4c is keg only, so it may be that it is not in the path. Check out my code in the arm build CI for discovering the position of the lib.
Also, does the linker complain at the end? And what is the type of the plugin?

martenrichter:
Can you please try if 0.1.0-macarmbuild.5 works on arm mac prebuild?
And also I would be interested if it is really an arm lib.
If it works, this PR should be reduced to the things, that fix local development build, since we have a nonunified solution for the prebuild arm problem.

achingbrain:
> Can you please try if 0.1.0-macarmbuild.5 works on arm mac prebuild?

Unfortunately it doesn't work for me:

```
load webtransport binary: ../build/Release/webtransport.node
dlopen(/Users/alex/Documents/Workspaces/libp2p/js-libp2p-webtransport/node_modules/@fails-components/webtransport/build/Release/webtransport.node, 0x0001): Library not loaded: /Users/runner/work/webtransport/armhomebrew/opt/icu4c/lib/libicuuc.71.dylib
  Referenced from: /Users/alex/Documents/Workspaces/libp2p/js-libp2p-webtransport/node_modules/@fails-components/webtransport/build/Release/webtransport.node
  Reason: tried: '/Users/runner/work/webtransport/armhomebrew/opt/icu4c/lib/libicuuc.71.dylib' (no such file), '/usr/local/lib/libicuuc.71.dylib' (no such file), '/usr/lib/libicuuc.71.dylib' (no such file)
```

Also the downloaded file is not a universal binary:

```console
% file node_modules/@fails-components/webtransport/build/Release/webtransport.node 
node_modules/@fails-components/webtransport/build/Release/webtransport.node: Mach-O 64-bit dynamically linked shared library arm64
```

This is the version:

```console
% npm ls @fails-components/webtransport
@libp2p/webtransport@1.0.3 /Users/alex/Documents/Workspaces/libp2p/js-libp2p-webtransport
└── @fails-components/webtransport@0.1.0-macarmbuild.5
```

achingbrain:
> I know, but you wrote that boring SSL does not work with universal binaries. Therefore I focussed on a cross compile for arm on intel.

Yeah, sorry - going a bit crazy. I think a universal binary is definitely needed once arm cross-compilation is sorted out, sometimes (because of tooling) you end up running x86_64 processes via rosetta on M1s which then need to load a x86_64 webtransport.node.  Other times you run it directly which needs a arm64 webtransport.node 🤪

Once both versions are built, stitching them together with `lipo` is straightforward, like in this PR.

> But this sounds like the lib version is simply not installed. Or that the build assumes a specific version.

I think it's because it's trying to load it from `/Users/runner/work/webtransport/armhomebrew/opt/icu4c/lib/libicuuc.71.dylib` which doesn't exist on my system since `runner` is the user in the GH actions container.

> Maybe you should run brew --prefix to see where it is installed.

```console
% brew --prefix
/usr/local
```

> Search for any libicuuc.*.dylib and check the version.

I have it at `/usr/local/Cellar/icu4c/71.1/lib/libicuuc.71.dylib ` and it's a x86_64 version:

```console
% file /usr/local/Cellar/icu4c/71.1/lib/libicuuc.71.dylib 
/usr/local/Cellar/icu4c/71.1/lib/libicuuc.71.dylib: Mach-O 64-bit dynamically linked shared library x86_64
```

martenrichter:
> > I know, but you wrote that boring SSL does not work with universal binaries. Therefore I focussed on a cross compile for arm on intel.
> 
> Yeah, sorry - going a bit crazy. I think a universal binary is definitely needed once arm cross-compilation is sorted out, sometimes (because of tooling) you end up running x86_64 processes via rosetta on M1s which then need to load a x86_64 webtransport.node. Other times you run it directly which needs a arm64 webtransport.node zany_face
> 
> Once both versions are built, stitching them together with `lipo` is straightforward, like in this PR.
But this PR does not solve the CI/CD pipeline with building it on intel mac. 
I need a receipi for this.

> 
> > But this sounds like the lib version is simply not installed. Or that the build assumes a specific version.
> 
> I think it's because it's trying to load it from `/Users/runner/work/webtransport/armhomebrew/opt/icu4c/lib/libicuuc.71.dylib` which doesn't exist on my system since `runner` is the user in the GH actions container.
> 
> > Maybe you should run brew --prefix to see where it is installed.
> 
> ```
> % brew --prefix
> /usr/local
> ```
> 
> > Search for any libicuuc.*.dylib and check the version.
> 
> I have it at `/usr/local/Cellar/icu4c/71.1/lib/libicuuc.71.dylib ` and it's a x86_64 version:
> 
> ```
> % file /usr/local/Cellar/icu4c/71.1/lib/libicuuc.71.dylib 
> /usr/local/Cellar/icu4c/71.1/lib/libicuuc.71.dylib: Mach-O 64-bit dynamically linked shared library x86_64
> ```
So it is not in path and the arm version is not installed, so it can not work.
Please install also the arm version via brew.
See for example:
https://earthly.dev/blog/homebrew-on-m1/

Btw. an alternative would be to fix the static icu build, which broke on mac.
So go back to:
https://github.com/fails-components/webtransport/commit/7b175907ed53acffc227cf26964104468b4045a7
fix the mac build problem in CMake.txt of the static link and then we go back to static. You may also want to check my attempts after this on 15 October, I went to the dylib you because I can only test mac via the CI/CD pipeline.
And then the libicu issue goes away.

martenrichter:
> I think it has to be static, no? Asking every user to install a bunch of homebrew deps doesn't seem realistic to me.
Yes, I assumed there would be a system lib that is compatible
> 
> I will see if I can help but as you might have guessed JS is my thing and not C++ laughing
And mine is more C++ ,but anyway, this is a matter of building and adding additional files to the icu static build.
So building it, finding out which files is missing and building it again, seeing what else is missing. And this does not work, if you have only CI/CD with no incremental builds (a commit and 40 minutes wait) so your plus is the local machine.

martenrichter:
May be this is enough
https://github.com/fails-components/webtransport/pull/120

martenrichter:
Ok now new attempt the version to test after building is `v0.1.0-macarmbuild.6`
Making it universal will be second step.
Even though do you have two node executables installed? Please check the type, if it is intel or mac, or both.

martenrichter:
I have check, and node itself at least provides separate binaries for intel and arm64. So I would guess it is not universal. Then of course the libs should also not be universal.  I also think that prebuild install dioes not support universal binaries..

achingbrain:
> Even though do you have two node executables installed? Please check the type, if it is intel or mac, or both.

```console
% which node
/Users/alex/.nvm/versions/node/v18.12.1/bin/node
% file /Users/alex/.nvm/versions/node/v18.12.1/bin/node
/Users/alex/.nvm/versions/node/v18.12.1/bin/node: Mach-O 64-bit executable arm64
```

Everything seems happy on the latest node 18, I'm going to close this for now, we can always revisit if necessary.

PR #121, opened 2022-11-10
Remember pending read when paused

davedoesdev:
Without this, because we can only check number of readable bytes, we miss a FIN that is received while paused.

You can see quiche regards having a FIN as notifiable here:

https://github.com/google/quiche/blob/main/quiche/quic/core/http/web_transport_stream_adapter.cc#L96

I see this happen occasionally when transferring lots of data and the stream ends while it's "paused"

### DIFF
--- a/src/http3wtstreamvisitor.cc
+++ b/src/http3wtstreamvisitor.cc
@@ -69,7 +69,10 @@ namespace quic
     void Http3WTStream::doCanRead()
     {
         if (pause_reading_)
+        {
+            can_read_pending_ = true;
             return; // back pressure folks!
+        }
         // first figure out if we have readable data
         size_t readable = stream_->ReadableBytes();
         // ok create a string obj to hold the data
--- a/src/http3wtstreamvisitor.h
+++ b/src/http3wtstreamvisitor.h
@@ -93,8 +93,9 @@ namespace quic
         void tryRead()
         {
             pause_reading_ = false;
-            if (stream_ && stream_->ReadableBytes() > 0)
+            if (stream_ && ((stream_->ReadableBytes() > 0) || can_read_pending_))
             {
+                can_read_pending_ = false;
                 doCanRead();
             }
         }
@@ -209,6 +210,7 @@ namespace quic
         bool fin_was_sent_ = false;
         bool stop_sending_received_ = false;
         bool pause_reading_ = false;
+        bool can_read_pending_ = false;
         std::deque<WChunks> chunks_;
     };
 
--- a/test/unidirectional-streams.spec.js
+++ b/test/unidirectional-streams.spec.js
@@ -125,4 +125,56 @@ describe('unidirectional streams', function () {
       'Did not receive the same bytes we sent'
     )
   })
+
+  it('handles fin when paused due to backpressure', async function () {
+    this.timeout(10000)
+
+    /** @type {Deferred<ReadableStream>} */
+    const serverStream = defer()
+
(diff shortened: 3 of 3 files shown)


martenrichter:
Thanks, I will look at it on the weekend.

davedoesdev:
Yes, I see `ReadableBytes() == 0`

martenrichter:
And what is the value of `fin_readable = sequencer_->IsClosed() && !fin_read_`?
Should we unblock if the fin arrives?

src/http3wtstreamvisitor.h
```diff
@@ -93,8 +93,9 @@ namespace quic
         void tryRead()
         {
             pause_reading_ = false;
-            if (stream_ && stream_->ReadableBytes() > 0)
```
martenrichter, line 96: May be add instead of `|| can_read_pending_` the check for a fin `|| (sequencer_->IsClosed() && !fin_read_)`.

src/http3wtstreamvisitor.h
```diff
@@ -93,8 +93,9 @@ namespace quic
...
+            if (stream_ && ((stream_->ReadableBytes() > 0) || can_read_pending_))
             {
+                can_read_pending_ = false;
```
martenrichter, line 98: Then this is not needed?

src/http3wtstreamvisitor.cc
```diff
@@ -69,7 +69,10 @@ namespace quic
     void Http3WTStream::doCanRead()
     {
         if (pause_reading_)
```
martenrichter, line 71: may be add `&& !(sequencer_->IsClosed() && !fin_read_)` but only if we want that a fin unblocks the stream.
Anyway for both, if you can provoke the problem, a test would be good.

src/http3wtstreamvisitor.cc
```diff
@@ -69,7 +69,10 @@ namespace quic
...
+        {
+            can_read_pending_ = true;
             return; // back pressure folks!
+        }
         // first figure out if we have readable data
         size_t readable = stream_->ReadableBytes();
         // ok create a string obj to hold the data
```
martenrichter, line 78: But of course a fin_read_ has to be added here somewhere.

davedoesdev:
I'll try to get a test when I get time, might be a few days.

davedoesdev:
Added a test, in `test/unidirectional-streams.spec.js`

martenrichter:
Ok thanks, after the tests have passed it will be merged.
I will then check if it is only a problem of a lonely FIN or indeed a problem connected to blocking.

martenrichter:
Thanks, I have not looked into it in detail. Your solution is the only one that works. (Under the week I had only to look into it, away from the development machine...., now I see that the fins are not exposed to the visitor)

davedoesdev:
Yes, fins aren't exposed.
Is there an ETA for a new release?

martenrichter:
I am waiting for achingbrain to confirm, that the arm binary is working. But I think I will release it anyway next weekend or before.
I usually follow the paradigm release often.

The question is, if we should ask upstream for exposing fins or if it is possible to work with a derived class without breaking too much other code.

davedoesdev:
How responsive is upstream (Google?)

martenrichter:
They usually answer queries in the WebTransport-dev google group or if you file a bug. And sometimes point me to the right portions of the code. But I never suggested a change to libquiche before, so I do not know.
Some issues regarding compatibility with windows are not fixed.

PR #130, opened 2022-11-30
test: run unit tests in chromium as well as node

achingbrain:
To ensure that the behaviour of the WebTransport client is the same in both node.js and Chromium, run the same unit tests in both environments.

This needed a bit of refactoring to do the server setup external to the tests.

1. The old test suite now lives in `old_test` to make the `test` dir a bit tidier
2. There are two new files in the test folder - `browser.js` and `node.js` that start the WebTransport server, shell out to Mocha or playwright-test
3. There were some differences in behaviour between `@fails-components/webtransport` and Chromium's `WebTransport` - I have changed the tests to expect the Chromium behaviour
4. `npm test` now runs node & browser tests, `npm run test:node` runs just node and `npm run test:chromium` runs tests on chromium
5. The browser tests are run as part of the linux test run, mostly because it's usually faster than Mac OS and Windows in GitHub actions
6. The `WebTransportSessionState` type was missing the `"connecting"` state
7. `Http3WTSession` instances now have the `"connecting"` state until ready, at which point they are `"connected"`
8. `WebTransportCloseInfo` instances with a `closeCode` of `0` used to have the `reason` `'closed'` - they now have an empty string as the `reason` to be consistent with Chromium

Because of 3. the tests are currently failing under node, I will push some extra commits to fix them.


martenrichter:
That PR is a great idea.
Once you identify the differences and the test succeeds, I will be happy to merge them.
Though I do not understand, the change in session, did I forget an await?

Btw during using the lib quite heavily in my project, triggered the development of this project,
 I have seen an occasional failure of client connections (I will fix it if I know, why). Probably a problem in the c++ client code, but maybe also an issue in my project. Though it would be interesting, where the difference to chrome is, maybe it is related?

achingbrain:
The tests sort of pass now.  I say sort of, because they pass for chromium reliably but under node they are quite flaky - streams sometimes don't receive data and sometimes random pauses occur - could there be some timing issues somewhere?

achingbrain:
Fixes #131

martenrichter:
> The tests sort of pass now. I say sort of, because they pass for chromium reliably but under node they are quite flaky - streams sometimes don't receive data 

Which kind of streams? Datagram or uni- and bidirectional? Datagrams are always unreliable per spec. 
Otherwise, it can be that data somewhere is blocked on the way.
I also had issues with the node event loop (where node's event loop was blocked, adding a microtask helped), but I thought these were gone.


> and sometimes random pauses occur - could there be some timing issues somewhere?

Yes and also a problem with the eventloop, I have the following ideas:
1.) It could be a stalled node eventloop. Try to add a setInterval with a delay of 1 ms. If then all random pauses are below this time, you have hit a problem in node itself.
2.) It can be a problem with the edge triggering 
in https://github.com/fails-components/webtransport/blob/84cd9d207de8f99fef7edf80dcafa7bcf13d3007/src/http3client.cc#L908
Tracking such bugs required placing prints of Date.now() and C++ equivalents all over the packet's path to find the origin of the wait. And I have also seen random failing client connections.
3.) Test if bigger Buffers help...

lib/client.js
```diff
@@ -35,36 +37,32 @@ export class Http3Client extends Http3WebTransport {
...
-      setTimeout(() => {
-        if (this.quicconnectedProm) {
-          this.quicconnectedProm.reject(new Error('Timeout client connection'))
-          delete this.quicconnectedProm
-        }
-      }, 8000)
-      await this.quicconnected
-      // now create Webtransport session
-      setTimeout(() => {
-        if (this.webtransportProm) {
```
martenrichter, line 58: Why did you drop the waiting for webtransport support. It can happen that is connects but never tell us, that it can have webtransport support.

martenrichter:
You wrote, that you use node18. Did you use node16 before?
Does it only fail with node18? (This question only applies, if problems with the event loop are present.)

lib/client.js
achingbrain: Because from what I can see, it doesn't wait for webtransport support, it just sets up the timer to reject the `webtransport` promise if webtransport support doesn't materialise.  The actual waiting happens in the `createWTSession` method so I moved the timer setup there so it's closer to the thing it's affecting.

From what I can see, these methods are only invoked by `lib/webtransport.js` and they are done one after another so I hope it's a bit easier to read now.

achingbrain:
> Which kind of streams? 

Unidirectional, though I think it might be a quirk of them getting run later in the test suite.  If you run the tests individually they pass.

The big difference here, I think, is that before this PR we'd set up and tear down servers for each test, now there's one server that services all of them, so it's doing more work which may have exposed an underlying timing or perhaps a cleanup issue.  Needs some investigating.

> Does it only fail with node18?

No, it's flaky with both node 16 and node 18.

lib/client.js
martenrichter: And still, fails, if we reach the time out and no webtransport support was signaled? (The lib only has an event telling that webtransport is possible, but not if it is unsupported). Then it is fine.

martenrichter:
> > Which kind of streams?
> 
> Unidirectional, though I think it might be a quirk of them getting run later in the test suite. If you run the tests individually they pass.

Then just flip the ordering of the tests and see if it changes....

> 
> The big difference here, I think, is that before this PR we'd set up and tear down servers for each test, now there's one server that services all of them, so it's doing more work which may have exposed an underlying timing or perhaps a cleanup issue. Needs some investigating.

Can be, then it may have to do with poll in the server side.
But add somewhere a setInterval with a timeout of 1 ms.
If it goes away, I know the reason.

> 
> > Does it only fail with node18?
> 
> No, it's flaky with both node 16 and node 18.
Ok, good.

achingbrain:
> But add somewhere a setInterval with a timeout of 1 ms.
If it goes away, I know the reason.

This doesn't seem to help, nor does changing the order.

I hope it's ok but I've added some more sophisticated logging in the form of the [debug](https://www.npmjs.com/package/debug) module.  Details are in the PR changes to the readme, but you can turn logging on and off for any or all components with the `DEBUG` environmental variable so no need to use `console.log` anywhere.  Specifying `DEBUG_TRACE` will show all the callback event data from C++ to JS.

lib/client.js
achingbrain: It does, but only because webtransport.js invokes one method after the other.  They could be one method, that might be a bit simpler?

achingbrain: It also could be made a bit more concise by using the [pTimeout](https://github.com/fails-components/webtransport/pull/130/files#diff-277293d1d50e356ece07e2607160b2de9f4f40ff0dc859222febe7588830266c) and [defer](https://github.com/fails-components/webtransport/blob/master/lib/utils.js#L5) functions (borrowed from [p-timeout](https://www.npmjs.com/package/p-timeout) and [p-defer](https://www.npmjs.com/package/p-defer) modules) but I think better to get the tests stable before making any more changes.

martenrichter:
There is some linting failing.... and ts is also complaining... do you see the result of the runners?

achingbrain:
For regular debugging:

```console
$ DEBUG=webtransport* npm run test:node
```

For trace debugging:

```console
$ DEBUG=webtransport* DEBUG_TRACE=true npm run test:node
```

lib/client.js
martenrichter: Yes one method is simpler, but it is harder to debug, what is failing on the c++ side, that is the reason it was split.

achingbrain: Fair enough - hopefully some more detailed logging on the js side will help there.

achingbrain:
Ok, so, I removed the really long timeouts so uses the mocha default of 2s, but sometimes the tests fail because of the random 5s delay that pops up.  I've restored the timeouts which stops them failing due to that.

Sometimes both a bidirectional and unidirectional test fails.  What seems to be happening is we [await on the result of closing the stream](https://github.com/fails-components/webtransport/pull/130/files#diff-2ae9f0537930a51ed77e5ee80c86f519c92e71e51875014508d73d0dfaaf8fceR19) on the client which creates a [pendingres](https://github.com/fails-components/webtransport/pull/130/files#diff-e56fabfb5e90fd8f6265cfbe84f3701a85261d884e198bf61de34958cee4864aR113) promise in `stream.js`.  Meanwhile as expected by the test behaviour the server closes the session.  Closing the session doesn't resolve the `pendingres` promise so the test hangs.

martenrichter:
Well spotted the hanging pendingres. But can you tell me which transaction hangs and remove the resolution?
Because this is a problem on the C++ side that needs to be fixed. The C++ side is not properly answering. So it needs to fail and if you tell me on which event it is waiting, I may be able to fix it.

achingbrain:
Ok great - I've removed the "fix" I applied.

> if you tell me on which event it is waiting

From the logs during a successful run, it's the `StreamNetworkFinish` event that isn't sent to the stream.

It also seems to get sent twice?

This is during the "sends data over an outgoing unidirectional stream" test:

Successful:

```
  webtransport:http3wtsession callback SessionReady +2ms
  webtransport:http3wtsession callback Http3WTStreamVisitor +1ms
  webtransport:http3wtstream callback StreamWrite +5s
  webtransport:http3wtstream callback StreamWrite +0ms
  webtransport:http3wtstream callback StreamWrite +0ms
  webtransport:http3wtstream callback StreamWrite +0ms
  webtransport:http3wtstream callback StreamWrite +1ms
  webtransport:http3wtstream callback StreamNetworkFinish +0ms
  webtransport:http3wtstream callback StreamNetworkFinish +0ms
  webtransport:http3wtsession callback SessionClose +1ms
  webtransport:http3wtsession onClose +0ms
```

Unsuccessful:

```
  webtransport:http3wtsession callback SessionReady +3ms
  webtransport:http3wtsession callback Http3WTStreamVisitor +0ms
  webtransport:http3wtstream callback StreamWrite +0ms
  webtransport:http3wtstream callback StreamWrite +0ms
  webtransport:http3wtstream callback StreamWrite +0ms
  webtransport:http3wtstream callback StreamWrite +0ms
  webtransport:http3wtsession callback SessionClose +2ms
  webtransport:http3wtsession onClose +0ms
```

When it's unsuccessful it looks like the session is closed while it's still in the process of sending data to the remote.

martenrichter:
Ok, that is some useful information.
But who is initiating the session close? Is it coming from the client side? (Can you add a log on the respective close called from js side? It may be that we have to wait somewhere in the tests or in the session object)

achingbrain:
> But who is initiating the session close? Is it coming from the client side? 

No - take a look at the ["sends data over an outgoing unidirectional stream" test](https://github.com/achingbrain/webtransport/blob/test/run-tests-in-browser/test/unidirectional-streams.spec.js#L26-L50) and the [corresponding server side stream handler](https://github.com/achingbrain/webtransport/blob/test/run-tests-in-browser/test/fixtures/server.js#L160-L180) you'll see it's the server that closes the session.

> It may be that we have to wait somewhere in the tests or in the session object

The test passes without issue when the client is Chromium so adding waits in the test doesn't sound right.  I think the client in node needs to close all of it's streams when the session ends, which includes resolving/rejecting any outstanding promises.

martenrichter:
> The test passes without issue when the client is Chromium so adding waits in the test doesn't sound right.

Agreed,  I commented in the code, that I was suspecting, that something was off.

>  I think the client in node needs to close all of it's streams when the session ends, which includes resolving/rejecting any outstanding promises.

No, the C++ side should cancel all these requests, because the answers may arrive later.
You debug output above, with 4 writestreams versus 5 writestreams says that something is different and there may be a problem on the C++ side.  So we may have to wait for the pending waits...
I have to look tomorrow closer into the code.

achingbrain:
> No, the C++ side should cancel all these requests, because the answers may arrive later.

I think that's what I meant, rather than the JS side.

Some other things, I noticed a C++ crash during [one of the test runs](https://github.com/fails-components/webtransport/actions/runs/3602434981/jobs/6069386393) though there doesn't seem to be a whole lot of info.

```
malloc_consolidate(): unaligned fastbin chunk detected
```

Also, in the actions queue quite a lot of the builds have stalled, it looks like uncaught exceptions can prevent the custom event loop shutting down in the old unit tests?  The jobs may need to be killed manually.

martenrichter:
> > No, the C++ side should cancel all these requests, because the answers may arrive later.
> 
> I think that's what I meant, rather than the JS side.
I am wondering, if the js side of the client close should wait for the pending writes in the streams, before sending the close.
I think it might be the right behavior for the network side. The spec is actually vague. Does the js side currently wait for the writes to complete, so that there is no pendingres in the end. Close before the writes finished might be the issue.

> 
> Some other things, I noticed a C++ crash during [one of the test runs](https://github.com/fails-components/webtransport/actions/runs/3602434981/jobs/6069386393) though there doesn't seem to be a whole lot of info.

This may be, but without a debug build and a core file... I can not do much.
> 
> ```
> malloc_consolidate(): unaligned fastbin chunk detected
> ```
> 
> Also, in the actions queue quite a lot of the builds have stalled, it looks like uncaught exceptions can prevent the custom event loop shutting down in the old unit tests? The jobs may need to be killed manually.

Ok I will do....

martenrichter:
Interesting, I see problems only with the connection of the client. The exactly the same issue that is also on my agenda, since my project also suffers from it. Definitely something on the C++ side.

martenrichter:
Something is messing with the stdout. I need to occasionally print things from C++ side for debugging..., what do I need to turn off?

martenrichter:
It is pretty weird. It seems that it takes roughly 8-9 seconds until data arrives from server to client (or until the poll is triggered ion the client side). Since it only happens on mac and Linux, it is probably an issue only on platforms using edge triggering.

martenrichter:
I still have no clue, so I stop investigating for this weekend.
If I run only one test at a time (commenting out the other tests) code, it runs fine.
So there is either a port not available.
Or either the server or the client has a problem with two clients with the same ip and (hopefully) a different port.
So may the port is not free or the older client is not shut down (it is always the second).
What I definitely saw, was that the paket took 10 seconds from sending on the service side (lowest layer) to the trigger of the poll. So problems with the polling in the event loop are also possible, but forcing level triggered does not help.

martenrichter:
Regarding logging. There are various errors from the quiche event loop, that are currently signaled by a printf (not possible to pass them to the JS side, different thread, so these errors might be lost). 
These need to be changed to another stream or std::out should work as before.

martenrichter:
Found it, it is an issue with upstream:
https://github.com/google/quiche/issues/29
commit with workaround is coming soon.

achingbrain:
Interesting.  On the logging thing, what's happening is the test runner spawns two child processes, one for the server and one for the test runner and attempts to merge the stdout and stderr of them both.  I've changed it to be a bit smarter and to wait for the output of a child process to contain a linebreak then only print up to the the linebreak, hopefully that will make the output better behaved.

martenrichter:
> Interesting. On the logging thing, what's happening is the test runner spawns two child processes, one for the server and one for the test runner and attempts to merge the stdout and stderr of them both. I've changed it to be a bit smarter and to wait for the output of a child process to contain a linebreak then only print up to the the linebreak, hopefully that will make the output better behaved.

Ok, but I got something like a javascript exception saying, that a JSON parser failed..., then I switched to stderr. But it may occur only during startup.... I will test it.

martenrichter:
The logging problems seem to be resolved.

martenrichter:
I do not understand, why are the checks passing. I did not commit the fix yet..., may be random...

achingbrain:
> Ok, but I got something like a javascript exception saying, that a JSON parser failed...

Ah right! The server process prints its cert hash and address when it starts and it’s read from stdout by the parent process to pass to the test runner.

This seems fine to me as this module shouldn’t be writing anything to stdout/stderr unless it’s a fatal error.

I can rethink the test setup if you disagree.

> I do not understand, why are the checks passing. I did not commit the fix yet...

Intermittent test failures are the best test failures 😂

martenrichter:
> > Ok, but I got something like a javascript exception saying, that a JSON parser failed...
> 
> Ah right! The server process prints its cert hash and address when it starts and it’s read from stdout by the parent process to pass to the test runner.
> 
> This seems fine to me as this module shouldn’t be writing anything to stdout/stderr unless it’s a fatal error.
> 
> I can rethink the test setup if you disagree.

Well if there is a timing issue or a segfault that I can not catch with the debugger. I usually rely on prints to stdout, but using stderr instead is also fine, I just have to know....

> 
> > I do not understand, why are the checks passing. I did not commit the fix yet...
> 
> Intermittent test failures are the best test failures 😂

At least in the logs, we see that the things that should fail, take almost 5 seconds. Maybe we got a faster runner.
Anyway fix is on its way, I just have to PR it to master.

martenrichter:
Fix is on its way to master
https://github.com/fails-components/webtransport/pull/134
https://github.com/fails-components/webtransport/pull/134/commits/b63dcd9af8e59b8677950ad7b2eb39e37e150e2e

martenrichter:
It is failing again but at a different test? Unfortunately, the CI logs are not conclusive

martenrichter:
It works on my development machine, so no clue, I will rerun the test.

martenrichter:
Also I am not sure, if we have to clear the outgoing write queue, the spec is vague.

martenrichter:
I also see the other issues, actually quite often. So good chance I can fix it next weekend.
Also for the other problem, I found a different cause after advise from upstream.

martenrichter:
For debugging, it would be good, if the pid is included in the output. So it is a bit difficult to identify, who is printing.

martenrichter:
Ok, today I will try to find the remaining issues. I see at least two issues. One sleeps with the event loop time out....

martenrichter:
Ok, I am not 100% sure, but most of the new logging stuff has to go.
The problem is, that js debug output is buffered. While debuging realtime timing issues I need that the messages are printed immediately from js and c++ and server and client, in order to see in which order they were called. In the moment it seems that the js side is delayed. That also means, that I am not making any progress until I revert parts of the changes back.

martenrichter:
Another issue, is if a function is called for getting the current time, it is not the actual time but a later time, since it is run in a later microtask. We need to use console.log or console.error directly.

martenrichter:
execa has also to go since it is also messing with stdout and stderr and this again makes debugging including the c++ side hard.
I try now to use spawn directly. Nevertheless I appreciate your work, just does not allow debugging in the current way...

martenrichter:
Instead of stdout for passing messages use node ipc:
https://nodejs.org/api/process.html#processsendmessage-sendhandle-options-callback
I will try to make it work with this and get debugging running.

achingbrain:
> We need to use console.log or console.error directly.

I think that might be the kind of thing you do locally to debug a certain problem but there's a bunch of good reasons not to to this - they are synchronous so can harm performance, this module will be used in environments where processes should be silent unless there's a fatal error, etc.

> js debug output is buffered

Is this when you run the tests or in general?  If it's when you run the tests, it's probably because yes, it's buffered - the runner sets up two processes, one for the server and one for the test runner and buffers until it sees a newline, this is to stop the output from the two processes interleaving.

Maybe it'd be better to just use `inherit` for `stdio` for the child processes and just live with the output occasionally getting a bit mangled?

> execa has also to go since it is also messing with stdout and stderr 

All execa does is add a few properties to the return from `child_process.spawn` and wraps it in a promise - 
https://github.com/sindresorhus/execa/blob/main/index.js#L84 so this doesn't sound right.

It does collect the output from stdout/stderr when you await on the promise result, but we're streaming the output straight to the output buffer so that sounds more like the cause of the problem.

> if a function is called for getting the current time, it is not the actual time but a later time, since it is run in a later microtask

Where is this happening?  If it's in a microtask it sounds like the function returns a promise rather than the current time?

martenrichter:
> > We need to use console.log or console.error directly.
> 
> I think that might be the kind of thing you do locally to debug a certain problem but there's a bunch of good reasons not to to this - they are synchronous so can harm performance, this module will be used in environments where processes should be silent unless there's a fatal error, etc.

Sure you are right. That is why I commented on the logs, when committing. 
So it has then to be sth like an 'if', using either your logging or `console.log` . But `console.log` has to be called directly.
I replaced the logger with a function (not async, but ()=>) calling `console.error` together with `Date.now()`, and I saw that it was called in the libuv loop iteration with its time stamp. But for the race condition, I am usually chasing I need the time from js and c++ to the ms accurate.

So I need sth, where I can quickly turn on the debugging output and find the problems your tests are identifying.


> 
> > js debug output is buffered
> 
> Is this when you run the tests or in general? If it's when you run the tests, it's probably because yes, it's buffered - the runner sets up two processes, one for the server and one for the test runner and buffers until it sees a newline, this is to stop the output from the two processes interleaving.

Yes, it is when I run the test. I did not try anything else. But especially when the tests are running, I need the tests to interleave (debug output is several 10000 lines....).
I am adding the pid to the lines to know, who is talking.

> 
> Maybe it'd be better to just use `inherit` for `stdio` for the child processes and just live with the output occasionally getting a bit mangled?

Yes, that is what I am trying.
> 
> > execa has also to go since it is also messing with stdout and stderr
> 
> All execa does is add a few properties to the return from `child_process.spawn` and wraps it in a promise - https://github.com/sindresorhus/execa/blob/main/index.js#L84 so this doesn't sound right.
> 
> It does collect the output from stdout/stderr when you await on the promise result, but we're streaming the output straight to the output buffer so that sounds more like the cause of the problem.

Yes it is the streaming, that adds a delay, and separates the outputs.

> 
> > if a function is called for getting the current time, it is not the actual time but a later time, since it is run in a later microtask
> 
> Where is this happening? If it's in a microtask it sounds like the function returns a promise rather than the current time?

No, that also happens without a promise. I think the reason is, that the callbacks are called from an external task. And I think the problem is limited to the callback.

martenrichter:
You are right execa is needed. I have now, a working debugging environment.
If you like, I can also commit my current working dir to an extra branch, so that you see, how I am tracking these errors. But it is messy.

martenrichter:
The first issue https://github.com/google/quiche/pull/30  solved, other problems are on this project side. I hope I found all, coming soon to master branch.

martenrichter:
I like to summarize the changes that I needed for debugging, before I delete them for pushing to main (they are not cleanup):

1. Use `stdio: ['inherit', 'inherit', 'inherit', 'ipc'] for execa for the server and `stdio: ['inherit', 'inherit', 'inherit']` for client and do not use OutputBuffers (or make it switchable)
2. Use IPC instead of stdout, here is working code:
 `if (process.send)
  process.send({
    address: `https://${address.host}:${address.port}`,
    certificate: certificate.fingerprint
  })
else console.error('No IPC channel')`
and the message handler on the other side.
3. The logging needs a switch to not invoke any buffers and use console.log. In case of the callback, debug code needs to be switchable (or commented), which uses directly `console.log` and relative - no intermediate functions call is allowed, since the libuv event loop will execute them in the next run of the callback and not when the c++ side is calling it.
4. If possible, pid and time in milliseconds since Unix epoch would be helpful, since these are also easily used on the c++ to identity a process.

martenrichter:
1. issue https://github.com/fails-components/webtransport/commit/f37de8d16c4306a80fc6481569301193b0383786 
2. issue https://github.com/fails-components/webtransport/commit/57883dd43152d582a69e17bdde4f22799c9a9b1d
3. issue https://github.com/fails-components/webtransport/commit/35747ca5f7899f7d0d9d88e309cfafea4549b458
will soon come to master.
Together with the upstream fix, I could not provoke after 20-30 runs a run exceeding 3 s.

martenrichter:
Master is merged, now we have to wait for upstream.

martenrichter:
Interesting linux browser test failed. I only tested node and of course one important fix is missing.

achingbrain:
The 5s delays seem to have gone, so that's great progress!

The `'handles fin when paused due to backpressure'` test fails consistently in the browser for me now - the server crashes on [this line](https://github.com/fails-components/webtransport/pull/130/files#diff-a3f5b1d31bbe70635388f5f4f489039180ad6b6427a2179bd876ba9188b035afR208).  A `0` is thrown by `await reader.read()` in that function, ie. a number, not an `Error` so there's no stack trace.

Is it related to the upstream fix?  Interesting that it doesn't break in node too, there must still be divergence in the client behaviours.

martenrichter:
> The 5s delays seem to have gone, so that's great progress!

Without the upstream patch? Then you are lucky, it was one of several causes, that caused the delay independently.
(5s is always a sign that the event loop is not woken up correctly... and I changed it for testing to 6 s)
> 
> The `'handles fin when paused due to backpressure'` test fails consistently in the browser for me now - the server crashes on [this line](https://github.com/fails-components/webtransport/pull/130/files#diff-a3f5b1d31bbe70635388f5f4f489039180ad6b6427a2179bd876ba9188b035afR208). A `0` is thrown by `await reader.read()` in that function, ie. a number, not an `Error` so there's no stack trace.

Mhh can be related to this one:
https://github.com/fails-components/webtransport/commit/35747ca5f7899f7d0d9d88e309cfafea4549b458
or this one
https://github.com/fails-components/webtransport/commit/f37de8d16c4306a80fc6481569301193b0383786
and maybe (but unlikely)
https://github.com/fails-components/webtransport/commit/dd70c48fc6f45695c109b0760084b53e9e2a1755
can revert them and tell me, which one it is. The browser stuff is not working for me.....(see below)

> 
> Is it related to the upstream fix? Interesting that it doesn't break in node too, there must still be divergence in the client behaviours.

I do not know... probably not, the upstream fix is tied to the 5-second delays, but it was one of several causes. The question is, is there a way to use a newer playwright version? The current one does not work with debian 11 inside my development container, and there were some changes to webtransport (mainly unrelated) recently) as well.
Anyway, I am splitting my time between my main project and this addon, and today is the main project time slot....

It can be a different behavior in the way the client closes the streams. You can place print out in the stream visitor onsomething methods and the calls with inform , to see how it differs.

martenrichter:
typescript checks fail in CI...

achingbrain:
> The question is, is there a way to use a newer playwright version? The current one does not work with debian 11 inside my development container, and there were some changes to webtransport (mainly unrelated) recently) as well.

It seems to work on the linux CI runner, so perhaps playwright isn't the problem - what's the error message you see?

martenrichter:
> > The question is, is there a way to use a newer playwright version? The current one does not work with debian 11 inside my development container, and there were some changes to webtransport (mainly unrelated) recently) as well.
> 
> It seems to work on the linux CI runner, so perhaps playwright isn't the problem - what's the error message you see?

The CI runner uses Ubuntu and I use debian bullseye in the container only a more recent version of playwright also run on debian bullseye. (I use the official node container, which is debian based.)

achingbrain:
Ok, fair enough - playwright-test now has the latest version of playwright-core, hopefully it'll run now.  I can only install Debian in a VM which then gives me arm64 linux so it's really hard for me to test it, sorry!

martenrichter:
Thanks, it runs now. So I can fix it at the weekend. It will probably not take long, I have just to see how chromium behaves and what the specs say about it. If the client differs or chromium or if it is undefined.

martenrichter:
Yesterday it did fail, but today it does not fail. Which is perfect for finding the error. Let's hope it changes on another day.

martenrichter:
Now I got an error. It says 'Connection lost' do you have the same? I am wondering if it is the client or the server and if the lib is missing keep-alive packets. Anyway time for today is up.

martenrichter:
Upstream included the patch
https://github.com/google/quiche/pull/30
test are running for updating master

test/unidirectional-streams.spec.js
```diff
@@ -1,180 +1,108 @@
...
     const writer = clientStream.getWriter()
 
-    for (const buf of input) {
+    for (const buf of KNOWN_BYTES) {
+      await writer.ready
       await writer.write(buf)
     }
 
-    await new Promise((resolve) => setTimeout(resolve, 1000))
+    await new Promise((resolve) => setTimeout(resolve, 500))
```
martenrichter, line 96: Increasing the timeout increases the likely hood of a Connection Lost or Session closed

test/fixtures/server.js
```diff
@@ -26,33 +32,223 @@ export async function createServer() {
...
+              const received = await readStream(stream, KNOWN_BYTES.length)
+
+              // if we did not get expected data, close the session with a reason
+              if (!ui8.equals(ui8.concat(KNOWN_BYTES), ui8.concat(received))) {
+                session.close({
+                  closeCode: 500,
+                  reason: 'data did not match'
+                })
+              } else {
+                session.close()
```
martenrichter, line 217: Adding a time out before this line reduces the likelyhood of failure

martenrichter:
I had a closer look at the failure of the last test.
The last test in the way it is written tests actually the client for a specific behavior, which is genuine to our implementation.
So it makes little sense to test in the current form of chromium, even though it is probably possible to revert the roles of client and server.

I have seen two failures at different points of communication. One was a 'Session closed' or `Communication lost`.
I could increase the failure rate by increasing delay on the client side. And I can reduce it by adding a delay on the server side-
I expect the following two factors, chromium cleans after a certain time the half-dead connections with only zombie streams up, or the session close from the server kills the session before the client waited for sending the fin on the stream (delay at server side). )May be trigger by IPC when to send it?)
The connection loss is very strange since I even do not see a connection attempt in the backend code.

So to conclude, I think, we should turn off this test for chromium, since it makes no sense in this correlation and I do not see really wrong behavior. If the delay, I could at least reduce the likely hood of the problems.

martenrichter:
Regarding the Connection lost, I see no Chlos coming in. So the browser does not seem to attempt to connect to the server???

martenrichter:
https://groups.google.com/a/chromium.org/g/web-transport-dev/c/bHEKE6J3RCk/m/3el4OUB8DgAJ?utm_medium=email&utm_source=footer

test/fixtures/server.js
achingbrain: > // time out is needed, since it can be received before the read is complete

I'm not sure what this means?

The second argument to [readStream](https://github.com/fails-components/webtransport/pull/130/commits/2541994ab96fd480f940ff6f492482b78520d186#diff-a3f5b1d31bbe70635388f5f4f489039180ad6b6427a2179bd876ba9188b035afR208) is how many bytes it should wait to read so if we've hit his branch, it means we read the correct number of bytes but they had the wrong content, so waiting an extra second shouldn't change anything.

test/unidirectional-streams.spec.js
```diff
@@ -93,16 +92,16 @@ describe('unidirectional streams', function () {
...
     await writer.close()
 
     // the remote will close the session cleanly if everything was ok
-    const result = await client.closed
+    await client.closed
 
-    // should receive the default close info
-    expect(result).to.have.property('reason', '')
-    expect(result).to.have.property('closeCode', 0)
+    // should receive the default close info, not true on chromium
```
achingbrain, line 103: > // should receive the default close info, not true on chromium

If the server receives the wrong bytes, it'll close the session with the `closeCode` `500` and the `reason` `'data did not match'` - by commenting out these assertions we don't guarantee that branch was not hit on the server.

test/unidirectional-streams.spec.js
```diff
@@ -93,16 +92,16 @@ describe('unidirectional streams', function () {
       await writer.write(buf)
     }
 
-    await new Promise((resolve) => setTimeout(resolve, 500))
+    await new Promise((resolve) => setTimeout(resolve, 2000))
```
achingbrain, line 95: The longer timeout shouldn't be necessary as after this we wait for the server to close the connection after it's done reading, rather than when the client thinks it's done writing.

test/unidirectional-streams.spec.js
martenrichter: > // time out is needed, since it can be received before the read is complete
Ups hard to remember, what I meant. I think it was the following.
Quic does not have head-of-line blocking, all streams and the session have a separate flow. So actually the session close can arrive before the read was complete. (I tracked the events down one by one) And this failed the test a couple of times since the read was intentionally blocked in this test. The delay ensures that the close arrives after the data so that the test can succeed. It is a time issue and did not occur by chance on the node tests.

martenrichter: > > // should receive the default close info, not true on chromium
> 
> If the server receives the wrong bytes, it'll close the session with the `closeCode` `500` and the `reason` `'data did not match'` - by commenting out these assertions we don't guarantee that branch was not hit on the server.

That is true. But the result on chromium was different as if it was not passed on, but it may interfere with the other change. I think I changed this first. Anyway, the CI is still failing this test, but I do not get a hint why. On my machine it is only occasionally failing with the connection loss problem, which I have no clue. (In a long-running test I also see quiche black hole detection..., no clue why)

achingbrain: > So actually the session close can arrive before the read was complete. (I tracked the events down one by one) 

Maybe I'm misunderstanding something here but are you sure this is possible?  In the test the client sends the bytes and waits for the session to end.  The server receives the bytes and then closes the session, either with an error code if the bytes were wrong or without if the bytes were correct.

As I see it the session cannot be closed before the read has completed because the server waits to read the expected number of bytes, e.g. it completes the read, before it closes the session.

I guess maybe somehow the client might receive the session close before it thinks it's finished writing the data?

> That is true. But the result on chromium was different as if it was not passed on, but it may interfere with the other change.

If we can't rely on the `closeCode`/`reason` fields this sounds like a bug?

> Anyway, the CI is still failing this test, but I do not get a hint why. 

Hmm, yeah - I had this failing reliably the other day, today it passes every time. :shakes fist:.

martenrichter: I do not remember everything in detail. It may as well be that the data was received but the fin came later and it was not the session. All I do recall, that it was a weird timing issue that arose since the data transfer was blocked but the closing not.
But anyway, this last unittest tests something internal in the client specific to the node implementation. So I do not think it makes sense at all to use it with chromium as client. 

I am not sure about closeCode, I may test it yourself with logging. But again we are testing Chromiums behavior here.

And yes, this not failing is a problem. On my machine, it really depends on whether or not the Backup process of my harddisk was running. More load more failures.

martenrichter:
I ported the new tests to the sister project, see PR and related commits https://github.com/fails-components/webtransport-ponyfill-websocket/pull/4 .    I found there one problem, related to the test flooding uncontrolled by the client. This may have triggered the problems, we have been seen with the browser. That the client buffer (which is not necessarily tied to the client) is flooded. Currently, it is pure speculation, since I could not reproduce it with my machine, but an automated test will show it. (See commit https://github.com/fails-components/webtransport/pull/130/commits/ec9f20fc203cda747f2e300b02a5b998b0603306 )

martenrichter:
At least for 3 runs the old problem seems to be gone.

test/fixtures/server.js
```diff
@@ -127,6 +127,7 @@ export async function createServer() {
                   try {
                     await writer.ready
                     await writer.write(Uint8Array.from([0, 1, 2, 3, 4]))
+                    await new Promise((resolve) => setTimeout(resolve, 50)) // do not flood everything
```
achingbrain, line 130: I don't know that this is a great solution - adding arbitrary timeouts in the test code implies there are race conditions hiding in the production code.

Furthermore it requires the sending peer to "do the right thing" which is almost certainly not to be the case in the real world - the receiving peer should be able to handle anything the client throws at it in order to project itself.

martenrichter: This is just a rate limitation. Without the delay, you are filling the client's buffer so full, that the client does not recover.  And it's actually unrealistic to flood with a network bandwidth available on the same pc. This is a substantial ddos attack in fact. And the problem is on the client side, it is actually an issue in chromium, so if something needs to be fixed it is on their side.

martenrichter: I have posted it as a bug to chromium:
https://bugs.chromium.org/p/chromium/issues/detail?id=1408176
may be the developer there can help.

martenrichter: Now it is crashing even with the change in the CI/CD.  It is really problematic, that it is so sporadic. But I still think it is caused by the datagrams and surfaces in the tests after the datagram test.

martenrichter:
It is the datagram, let check, if it is the client to server part or both.

martenrichter:
The situation is a bit blurry, but it seems, that the session is not properly closed in one of the datagram case. But we have to wait for more precise debugging code.

martenrichter:
Since last week it is not failing any more in CI/CD... reason unknown, may be different quiche version helped or different load on CI/CD servers....

martenrichter:
Or the debug messages change the timing....

martenrichter:
For a long time, the tests did not fail in CI/CD. Today the tests failed again. I have now committed a change that the tests are run with separate chromium instances. This is to answer the query of a chromium developer.

martenrichter:
Ok it fails also with restarted clients, that means it is a problem in the server.

martenrichter:
It seems, that I forgot to include the datagram test, but it failed anyway. So then it was just the unreliable occurrence of error, that lead to the conclusion. Anyway, since it fails at the creation of the session, it must be still a previous connection.

martenrichter:
All attempts so far failed, and since it does not fail on my development machine, it is relatively hopeless to find the cause. Waiting after a change for committing and  20 Minutes of compile time in CI brings me nowhere. I will try next weekend if I can get somewhere a Vm or something like GitHub codespaces, where I can hopefully reproduce it and develop and edit on the same machine.

martenrichter:
Do you know by any chance how to add to the playwrights the capturing of a chromium netlogs: https://support.google.com/chrome/a/answer/6271171?hl=en This is what upstream needs to help and I am not familiar, if we can add this to playwright tests...

martenrichter:
I was able to get a netlog, by copying all Browser side test code to a file and running it in a browser by hand. Let's see what upstreams say. I have also found a bug in the fixtures/server

martenrichter:
The Netlog was reviewed by upstream, but the connection that failed was missing. Moreover, the logging on the browser side stops exactly when the problems occur, which according to upstream is something, upstream had not seen. So it is now up to upstream, it is clearly weird.

martenrichter:
So question is, if we want to make the browser test optional and proceed with the merge? Or run it several times and it is enough if it works once and change this once upstream is ready? Otherwise, this may be blocked by a problem in chromium for a while.

martenrichter:
Of course, there would be another way. If every test is run in a separate chromium, there should be no problems and it will test the functionality of the server anyway. In this way, we will not have to wait for the chromium side.

martenrichter:
Update: I found some time to turn on a debugger for Chrome. All I can see is, that I see a problem on the browser side. So it is not a fault of the library, so it will never work with the current browser version.

martenrichter:
Maybe some more details. It is a seg fault in the network service of chromium, that occurs in a specific timing of network messages.  I documented the details in the bug report for chromium. The delay times between the tests helped because the browser had time to restart the network service within the delay time. So for the purpose of the library, it can serve as a workaround until the problem in chromium itself is fixed. Ironically the problem is very similar to a bug, we had previously in the library, which lead to the last test. This explains, why it triggers the problem so reliably.

martenrichter:
Ok, after in 
https://bugs.chromium.org/p/chromium/issues/detail?id=1408176
chrome was patched. Chrome canary now does not fail on the most likely fail ( version115.0.5763.0).
So once this is in stable, one can ask the test tool manufacturer to update, and then we can continue to get this going.
However, it still fails occasionally on test 'bidirectional_server_initiated_echo', however, I believe it may be:
https://bugs.chromium.org/p/chromium/issues/detail?id=1427311&q=component%3ABlink%3ENetwork%3EWebTransport&can=2
At least the behavior on the server side looks the same as the reported pattern.
I will try, if can find this hopefully last one out.

martenrichter:
Ok, the last bug is not on chrome side, but I false construction of the tests. I will comment in the source code.

test/fixtures/server.js
martenrichter: Wait this is wrong. KNOWN_BYTES length is actually the number of chunks in KNOWN_BYTES .
As WebTransport does not gurantee, that the chunk are not merge or split, this cause the test occasionally to fail. So actually the number of bytes inside the chunks of KNOWN_BYTES added should be passed.

test/bidirectional-streams.spec.js
```diff
@@ -1,138 +1,86 @@
...
-    ]
-
     const stream = await client.createBidirectionalStream()
-    await writeStream(stream.writable, input)
+    await writeStream(stream.writable, KNOWN_BYTES)
 
-    const output = await readStream(stream.readable)
-    expect(output).to.deep.equal(
-      input,
+    const output = await readStream(stream.readable, KNOWN_BYTES.length)
```
martenrichter, line 50: KNOWN_BYTES.length is the number of chunks and this causes the test occasionally to fail.
Anyway, and this probaly a bug in the browser, the browser sometimes does not send a fin. So if the browser behaves correctly this should not be an issue, but the browser does not.

test/fixtures/read-stream.js
martenrichter, line 1: The problem is here. Expected is the number of chunks. But the chunk number is not guaranteed. So this test fails, if the chunks are merged or splitter, which is permitted per spec and for higher data rates normal!

martenrichter:
So I have asked playwright-test to update to the new playwright-core version without the webtransport chromium problem. After this hopefully, this will taking off. And I will see which issues remain and are not caused by the chromium network service segfault.

martenrichter:
One issue seems to remain. I will try to track it down at the weekend. Let's hope it is on the node side and not on the browser side. (There are other bugs reported in chromium about flaky tests in the browser, so it is possible).

martenrichter:
It is only failing in github codespaces and not on Docker container inside Windows and not on github codespaces. I will now try if it is an issue with ubuntu, on which the github tests are running. So temporarily switching the development container.

martenrichter:
Or it is the node version, so I will switch to node:18 first

martenrichter:
The last fails only happen on release builds...., timing issue?

martenrichter:
I ask upstream for the new issue, if they can take a look and tell me, if it is a server fault or chromium problem.
I do not want to spend a lot of time on my side to discover, that it is a chromium issue in the end:
https://bugs.chromium.org/p/chromium/issues/detail?id=1455664

martenrichter:
I have created a setup, where I can build and debug chrome. 
Maybe it was not a good idea to do this on my hdd. Anyway, I am coming closer to the cause of the problem.

It seems that resetstream or cancelstream are not always passed by quiche to the javascript stream, when a session is closed. It seems to depend on the calling sequence. 
I have seen this before on the node plugin, I thought it was my fault, that I have to reset the streams at the javascript side manually, but in fact quiche should do this.
So my code is currently working around a quiche bug. I will remove this workaround now, and try to find the root cause in libquiche and hopefully, then submit a PR to upstream. At least this is my plan.

martenrichter:
Ok, I found the cause again inside chromium:
https://bugs.chromium.org/p/chromium/issues/detail?id=1455664&q=webtransport&can=2
I hope that upstream will include my fix.
Until then and a subsequent release of the fixed chromium, we have to wait for the browser test to pass.
I am sorry for the delay, but again not my bug.

martenrichter:
Update, I have add a CL for Chromium:
https://chromium-review.googlesource.com/c/chromium/src/+/4697154
as it will take again several month to end up in the release version for testing, I have added some workarounds in the code.
I think, this is ok, since I know precisely, what is failing on the browser side. These exceptions will be removed once Chromium is fixed, but I think we can move ahead with this PR in the meantime. (I will check, if I can remove some cancels at the server side, which workaround the bug, before I knew details.)
@achingbrain Do you agree?

martenrichter:
https://chromium.googlesource.com/chromium/src/+/d6b544ac01b52381a82fc88d3a28d27642c61cd0
fix landed in chromium main

martenrichter:
Add release tracker for last patch in chromium:
https://chromiumdash.appspot.com/commit/d6b544ac01b52381a82fc88d3a28d27642c61cd0
so that we can see when it hits canary and dev etc.

martenrichter:
I ask again playwright test for an update, as the patch is now in stable.

martenrichter:
Playwright-tests is updated, let's see if the tests are finally stable....

martenrichter:
```

[playwright-test] Autodetected "mocha" as the runner.
TypeError: crypto.getRandomValues is not a function
    at fillPool (file:///home/runner/work/webtransport/webtransport/node_modules/playwright-test/node_modules/nanoid/index.js:9:12)
    at nanoid (file:///home/runner/work/webtransport/webtransport/node_modules/playwright-test/node_modules/nanoid/index.js:40:3)
    at new Runner (file:///home/runner/work/webtransport/webtransport/node_modules/playwright-test/src/runner.js:49:48)
    at file:///home/runner/work/webtransport/webtransport/node_modules/playwright-test/cli.js:310:18
```
Bad luck, the test harness seems to be broken...

martenrichter:
It works now. (As the test includes 10 browser tests).
@achingbrain: Maybe you should have a final look over the current state (then say ready for review), but I think merging soon, should not be a problem,

martenrichter:
I merge it know anyway..., as all issues seems to be addresses without workarounds.

PR #132, opened 2022-11-30
fix: pin typescript to 4.8.4

achingbrain:
A recent release of TypeScript has updated the types to make the `Readable` and `Writable` generics a union with `ArrayBuffer`.  This is a mistake and breaks the build, so pin TypeScript to an earlier version that works.


achingbrain:
Ah, the build in master is still failing even after this change because it's taking the `typescript` version from the `package-lock.json` file instead of `package.json`.  It needs regenerating or removing.

Issue #233, opened 2023-12-30
Types missing on version 1.0.4


endel:
Hi @martenrichter, congrats on the latest update 👏  I've just tried it out and noticed TypeScript type definitions are missing, though:

<img width="742" alt="Screenshot 2023-12-30 at 10 32 44" src="https://github.com/fails-components/webtransport/assets/130494/19b12b83-0500-4dea-b195-3a2a182c7df3">

Cheers!

martenrichter:
Ok, I was so happy, that the compiler did not throw any errors at me. 
I hope, I have fixed it, in the branch for dealing with firefox unit tests.
Sorry for this, but the problem with typescript is that I do not use it so it is very likely to break.
You can test if it makes sense for you. I may release a new bugfix release this year.

PR #236, opened 2023-12-31
fix generating TypeScript type definitions

endel:
Hi @martenrichter, 

I've made some changes to `tsconfig.json` and changed to generate type definitions only before publishing (`prepublishOnly`)

I'm still getting used to the codebase, let me know if I'm missing something.

### Before

Screenshot 2023-12-31 at 10 32 22
Screenshot 2023-12-31 at 10 32 07

### After

image

image

---

All the best and happy new year ✨

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,6 @@
     "test:webkit:http2:ponyfill": "npm run test:webkit:http2:ponyfill --workspace main",
     "oldtest": "npm run oldtest --workspace main",
     "oldtesthttp2": "npm run oldtesthttp2 --workspace main",
-    "types": "tsc --build",
     "lint": "npm run lint --workspaces",
     "prebuild": "npm run prebuild --workspace transports/http3-quiche",
     "build-debug": "npm run build-debug --workspace transports/http3-quiche",
--- a/package.json
+++ b/package.json
@@ -30,6 +30,7 @@
     "test:webkit:http2:ponyfill": "npm run test:webkit:http2:ponyfill --workspace main",
     "oldtest": "npm run oldtest --workspace main",
     "oldtesthttp2": "npm run oldtesthttp2 --workspace main",
+    "types": "tsc --build",
     "lint": "npm run lint --workspaces",
     "prebuild": "npm run prebuild --workspace transports/http3-quiche",
     "build-debug": "npm run build-debug --workspace transports/http3-quiche",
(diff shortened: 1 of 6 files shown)


martenrichter:
Also happy new year to you! (In a couple of hours I suppose).
Regarding the change, I actually would like to keep the "types" command, so that I can verify, that the types are ok, before I hit 
The rest I will comment within the commit.

package.json
```diff
@@ -27,7 +27,6 @@
     "test:firefox:http2:ponyfill": "npm run test:firefox:http2:ponyfill --workspace main",
     "oldtest": "npm run oldtest --workspace main",
     "oldtesthttp2": "npm run oldtesthttp2 --workspace main",
-    "types": "tsc --build",
```
martenrichter, line 30: Does this generate a problem? The version release today, does it still have problem with types?

transports/http3-quiche/package.json
```diff
@@ -20,7 +20,8 @@
     "rebuild-debug": "node build.js rebuild-debug -D",
     "build-debug": "node build.js build-debug",
     "prebuild": "node build.js prebuild",
-    "lint": "eslint lib"
+    "lint": "eslint lib",
+    "prepublishOnly": "tsc"
```
martenrichter, line 24: Please look also at the github action for the actual publishing...

package.json
endel: Just checked the version of today (`1.0.8`) and https://github.com/fails-components/webtransport/issues/233 still persists due to the path to `./dist/lib/index.d.ts` that is failing. 

A potential solution could be pointing `"types": "./dist/main/lib/index.d.ts"`, but I believe having both projects compiled under `dist` was not intentional. 

Below is a screenshot of both modules (`1.0.8`) after being installed via NPM (`@fails-components/webtransport` and `@fails-components/webtransport-transport-http3-quiche`), you can see both projects have the same output under `dist` folder, whereas they should only contain their own output:

Screenshot 2023-12-31 at 12 43 12

martenrichter: Ok, the screenshots match my development environment.
Yes, it was not intentional.
But I had to trouble with the crossreferencing of the two packages in types.
Does a run of `tsc --build` at workspace level run fine with your changes?

endel:
Thanks for the review, I've just updated bringing back the `"types"` script and removing the `prepublishOnly`, I believe publishing should work as before now 🙌

martenrichter:
Ok, I will then hit the CI/CD jobs to see if they run, and maybe merge it later. I will probably wait for a release (or do the types not work at all?) until another issue pops up.

PR #254, opened 2024-02-28
docs: fix homepage link in package.json

achingbrain:
The homepage link on the [npm page](https://www.npmjs.com/package/@fails-components/webtransport) for this module is broken, so update it to point to the right place.


martenrichter:
Thanks!

PR #256, opened 2024-03-03
Use peek readable to fix receiving fins?

martenrichter:


transports/http3-quiche/src/http3wtstreamvisitor.cc
```diff
@@ -142,8 +154,10 @@ namespace quic
         if (send_fin_)
         {
             bool success = stream_->SendFin();
-            if (success)
+            if (success) {
                 fin_was_sent_ = true;
+                getJS()->processStreamNetworkFinish(NetworkTask::streamFinal);
+            }
```
SgtPooki, line 160: in what cases will `success == false` ? should we do anything with that case?

martenrichter: We will retry the next round, as the network was probably blocked.
All other failure cases have separate message handler, that inform js land.

PR #257, opened 2024-03-04
fix: check closed status of socket during callbacks

achingbrain:
Sometimes calling `getSendQueueCount` can cause an internal crash in the `dgram` module.

I think it may be because packets are being sent during server shutdown - guarding on the `this.closed` property similar to the `sendPacket` method avoids the crash.

```
TypeError: Cannot read properties of null (reading 'getSendQueueCount')
    at Socket.getSendQueueCount (node:dgram:995:36)
    at Http3WebTransportClientSocket.packetSendCB (file:///Users/alex/Documents/Workspaces/libp2p/js-libp2p-webtransport-listener/node_modules/@fails-components/webtransport-transport-http3-quiche/lib/socket.js:118:24)
    at processTicksAndRejections (node:internal/process/task_queues:82:21)
```

### DIFF
--- a/transports/http3-quiche/lib/socket.js
+++ b/transports/http3-quiche/lib/socket.js
@@ -85,6 +85,10 @@ export class Http3WebTransportSocket {
   }
 
   doProcessBufferedChlos() {
+    if (this.closed) {
+      return
+    }
+
     this.cobj.processBufferedChlos()
     this.chlosSched = false
   }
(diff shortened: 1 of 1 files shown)


martenrichter:
Sure, I have also seen this. But it was during another problem, so I thought it was a side effect.

martenrichter:
Sorry I did not think this through, I reverted the PR.
This may cause promises to be not resolved.
The protection must to be only in the check, if writes are blocked.

PR #262, opened 2024-03-05
Fix getSendQueueCount (modified change org by achingbrain)

martenrichter:


martenrichter:
The same as
https://github.com/fails-components/webtransport/pull/257
but without preventing a call to process buffered chlos.

PR #263, opened 2024-03-05
fix: test that quicheLogVerbose is a number

achingbrain:
According to the [types](https://github.com/fails-components/webtransport/blob/master/main/lib/types.ts#L272) `quicheLogVerbose` when passed as an option should be a number, the test here asserts it's a function, so it's not possible to enable verbose quiche logging.

### DIFF
--- a/transports/http3-quiche/src/librarymain.cc
+++ b/transports/http3-quiche/src/librarymain.cc
@@ -38,7 +38,7 @@ namespace quic
       }
 
       quiche_cmd_line.push_back(std::string("-v"));
-      if (lobj.Has("quicheLogVerbose") && lobj.Get("quicheLogVerbose").IsFunction())
+      if (lobj.Has("quicheLogVerbose") && lobj.Get("quicheLogVerbose").IsNumber())
       {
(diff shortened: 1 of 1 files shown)


achingbrain:
@martenrichter any idea what the number should be to get quiche to output something?

martenrichter:
If I remember correctly, an integer 1,2,3 for the different levels. 
But I think it is coming through stderr, stdout. I can not remember if I programmed a redirect to the javascript console, I wanted.

achingbrain:
Hmm, I still don't see anything in the terminal - I can see the output of `printf` lines I put in the c++ code, so I would have thought I'd see some quiche log output if this option does what it's supposed to do.

martenrichter:
It may be stderr, you have to look at the absl logging impl.

achingbrain:
I don't think stderr would be hidden, I'm running things using the node cli in zsh, stdout and stderr appear as expected but nothing from quiche.

martenrichter:
Then I do not know, it may be broken. So, I may have used it before the reorganization. I can look into it..., but it does not seem that I redirected the log output... Most times I use print for diagnosis for the stuff that I need.

PR #264, opened 2024-03-05
fix: constrain quicheLogVerbose value

achingbrain:
The valid values [are](https://fuchsia.googlesource.com/third_party/abseil-cpp/+/refs/tags/20230125.3/absl/base/log_severity.h#69) -1 | 0 | 1 | 2 | 3 so cause a type error if anything else is passed.

### DIFF
--- a/main/lib/types.ts
+++ b/main/lib/types.ts
@@ -266,10 +266,17 @@ export interface WebTransportSessionImpl extends WebTransportSession {
   state: WebTransportSessionState
 }
 
+export type QUICHE_LOG_OFF = -1
+export type QUICHE_LOG_INFO = 0
+export type QUICHE_LOG_WARNING = 1
+export type QUICHE_LOG_ERROR = 2
+export type QUICHE_LOG_FATAL = 3
+export type QUICHE_LOG = QUICHE_LOG_OFF | QUICHE_LOG_INFO | QUICHE_LOG_WARNING | QUICHE_LOG_ERROR | QUICHE_LOG_FATAL
+
 export interface HttpWebTransportInit extends WebTransportOptions {
   host: string
   port: string | number
-  quicheLogVerbose?: 1 | 2 | 3
+  quicheLogVerbose?: QUICHE_LOG
   forceIpv6?: boolean
   localPort?: number
 }
--- a/main/lib/types.ts
+++ b/main/lib/types.ts
@@ -269,7 +269,7 @@ export interface WebTransportSessionImpl extends WebTransportSession {
 export interface HttpWebTransportInit extends WebTransportOptions {
   host: string
   port: string | number
-  quicheLogVerbose?: number
+  quicheLogVerbose?: 1 | 2 | 3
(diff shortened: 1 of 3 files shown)


achingbrain:
It might be these ones? https://fuchsia.googlesource.com/third_party/abseil-cpp/+/refs/tags/20230125.3/absl/base/log_severity.h#69

martenrichter:
Yes, this must be the one, and -1 is off. It is the default.

achingbrain:
The types here are hard to use because the `WebTransportBase` constructor [takes a WebTransportOptions](https://github.com/fails-components/webtransport/blob/master/main/lib/webtransportbase.js#L26) rather than a [HttpWebTransportInit](https://github.com/fails-components/webtransport/blob/master/main/lib/types.ts#L269) so the `quicheLogVerbose` isn't a valid option.

It needs to be added to the [WebTransportOptions interface](https://github.com/fails-components/webtransport/blob/master/main/lib/dom.ts#L74) which I see already has some non-standard options defined in it.

FWIW most of `dom.js` can be removed now since the `WebTransport` types are defined in TypeScript's DOM typings which they weren't back when `dom.js` was added.

martenrichter:
> The types here are hard to use because the `WebTransportBase` constructor [takes a WebTransportOptions](https://github.com/fails-components/webtransport/blob/master/main/lib/webtransportbase.js#L26) rather than a [HttpWebTransportInit](https://github.com/fails-components/webtransport/blob/master/main/lib/types.ts#L269) so the `quicheLogVerbose` isn't a valid option.

So far, I did not consider it a public interface..., but feel free to add it in a pact.

> 
> It needs to be added to the [WebTransportOptions interface](https://github.com/fails-components/webtransport/blob/master/main/lib/dom.ts#L74) which I see already has some non-standard options defined in it.
> 
> FWIW most of `dom.js` can be removed now since the `WebTransport` types are defined in TypeScript's DOM typings which they weren't back when `dom.js` was added.

I do not think so. The WebTransport spec changes very often, so I would have missing properties. So it is better to keep it, as it will for sure miss some of the new options.

achingbrain:
In that case it might be better to extend the built-in types to add new properties rather than redefining them?

Otherwise you can end up in all sorts of trouble when some field three levels down in an interface has a type that's missing a `null` or some other trivial thing.

martenrichter:
> Otherwise you can end up in all sorts of trouble when some field three levels down in an interface has a type that's missing a `null` or some other trivial thing.
Precisely, this happens already with the Streams, where the definition from node.js is slightly different from the dom one.
And as this is the only project, I come in touch with types, it is taking too much time.

achingbrain:
The Node.js types are quite tedious, yes.  They are written by the community though, sometimes you have to PR [DefinitelyTyped/DefinitelyTyped/types/node](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/node), but the maintainers there are at least pretty good about merging PRs quickly.

PR #267, opened 2024-03-08
Fix http3 client

martenrichter:


martenrichter:
I have to test this in a more complex app before merging.

PR #280, opened 2024-04-03
feat: support query strings

### LINKED ISSUE #279: Support query strings?
If I connect to a WebTransport server from node using the `WebTransport` implementation from this module with a query:

```js
const wt = new WebTransport('http://example.com/session_with_query?foo=bar')
```

..the server does not receive the `?foo=bar' part of the URL, only the `/session_with_query` part which means you can't connect to servers that expect/require query args.

It seems to be because only the `.pathname` property of the `URL` is used:

https://github.com/fails-components/webtransport/blob/b7ef52877546f7b0597ba31b41d70554bf8728e6/main/lib/webtransport.node.js#L80

If it's changed to:

```js
.then(() => client.createWTSession(sessionint, `${ourl.pathname}${ourl.search ?? ''}`))
```

..then the remote receives the query string.

----

From the server perspective, if the client with the patch above connects to the URL with the query string, this server session read stream does not yield the session:

```js
for await (const session of getReaderStream(
  server.sessionStream('/session_with_query')
)) {
  // handle session
}
```

...instead you have to register a path that includes the query:

```js
for await (const session of getReaderStream(
  server.sessionStream('/session_with_query?foo=bar')
)) {
  // handle session
}
```

This isn't very flexible for obvious reasons.

Could the WebTransport Session on the server side have a `.url` property [similar to message.url from `http`](https://nodejs.org/api/http.html#messageurl) or [request.url from `http2`](https://nodejs.org/api/http2.html#requesturl)?

Then it could be up to the js server to parse the query string and handle the request.

I'd started working on a PR but it seems we keep a list of registered paths in [js](https://github.com/fails-components/webtransport/blob/master/main/lib/server.js#L131) and in [c++](https://github.com/fails-components/webtransport/blob/master/transports/http3-quiche/src/http3serverbackend.h#L91) so it wasn't as straightforward as I had hoped.

achingbrain:
Adds support for creating WebTransport session with URLs that include query strings:

```js
const wt = new WebTransport('https://example.com/some-endpoint?foo=bar')
```

...and also accessing the path via the `.url` property which is consistent with Node http/http2 requests

```js
const sessionReader = server.sessionStream('/some-endpoint').getReader()
const { value: session } = await sessionReader.read()

console.info(session.url)
// /some-endpoint?foo=bar
```

Fixes #279

### DIFF
--- a/main/lib/server.js
+++ b/main/lib/server.js
@@ -1,5 +1,5 @@
 import { ReadableStream } from 'node:stream/web'
-import { HttpWTSession } from './session.js'
+import { HttpWTServerSession } from './session.js'
 import { WebTransportError } from './error.js'
 import { Http2WebTransportServer } from './http2/node/index.js'
 import { isIPv4 } from 'net'
@@ -51,7 +51,7 @@ const quicheLoaded = new Promise((resolve, reject) => {
 })
 
 /**
- * @typedef {import('./types').WebTransportSessionImpl} WebTransportSession
+ * @typedef {import('./types').WebTransportServerSessionImpl} WebTransportSession
  * @typedef {import('./types').NativeHttpWTSession} NativeHttpWTSession
  * @typedef {import('./types').HttpServerEventHandler} HttpServerEventHandler
  * @typedef {import('./types').HttpWTServerSessionVisitorEvent} HttpWTServerSessionVisitorEvent
@@ -130,7 +130,7 @@ export class HttpServer {
     /** @type {Record<string, ReadableStream>} */
     this.sessionStreams = {}
 
-    /** @type {Record<string, ReadableStreamDefaultController<HttpWTSession>>} */
+    /** @type {Record<string, ReadableStreamDefaultController<HttpWTServerSession>>} */
     this.sessionController = {}
 
     this.port = null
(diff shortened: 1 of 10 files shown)


main/lib/server.js
```diff
@@ -268,15 +268,19 @@ export class HttpServer {
    * @param {HttpWTServerSessionVisitorEvent} args
    */
   onHttpWTSessionVisitor(args) {
+    // match session streams on path without query string
+    const path = args.path.split('?')[0]
```
martenrichter, line 272: As this is not the header path field, we should not temper here. So please remove.

martenrichter:
No, I do not think I will include the change. It is too specific for your application case. Please see the callback mechanism, which is very flexible for many use cases. And I will be happy to answer questions if the mechanism is unclear. (Especially as I assume, that you want to deny requests early, that are not authorized.)

main/lib/server.js
achingbrain: The `args` passed here during the new `should connect to a path with a query in the URL` test are:

```js
{
  session: Http3WTSessionVisitor {},
  path: '/session_with_query?foo=bar'
}
```

If we don't remove the query string from the path, the session is never passed to the session stream controller for the endpoint and the test fails.

martenrichter:
But you can do it already with the other mechanism.., which is transport independent, as a middleware in express would handle it. I will post to the issue an untested example.

PR #282, opened 2024-04-05
test: add test for user data and request callbacks

achingbrain:
Adds a request callback that adds any search query to the user data object.

Adds a test that stringifies the user data object and sends it to the client, then verifies it on the client side.


martenrichter:
lgtm

PR #293, opened 2024-05-20
Replace assert with with

### LINKED ISSUE #91: Install fails on LTS node
The `build.js` file uses import assertions, these fail on the present LTS node (16.x):

```console
% npm install @fails-components/webtransport
npm ERR! import pkg from './package.json' assert { type: 'json' }
npm ERR!                                  ^^^^^^
npm ERR! 
npm ERR! SyntaxError: Unexpected identifier
npm ERR!     at ESMLoader.moduleStrategy (node:internal/modules/esm/translators:139:18)
npm ERR!     at ESMLoader.moduleProvider (node:internal/modules/esm/loader:236:14)
npm ERR!     at async link (node:internal/modules/esm/module_job:67:21)
```

Node 16 is still a year away from end of life, is there any chance this can be refactored to work on node 16?

martenrichter:
Closes #91


main/test/fixtures/server.js
```diff
@@ -243,6 +242,7 @@ export async function createServer() {
                 }
                 await getReaderValue(session.incomingUnidirectionalStreams)
                 await session.close()
+                // eslint-disable-next-line no-unused-vars
               } catch (error) {
                 // do not crash server, if a problem occurs...
               }
```
achingbrain, line 248: If you want to ignore the error you can just omit it, then there's no need to disable linting:

```js
} catch {
  // do not crash server, if a problem occurs...
}
```

Issue #365, opened 2025-04-18
Writing new Uint8Array() causes server freeze with quic_bug_10586_6: span.empty() && !fin


guest271314:
During testing. 

Server. `bun build --target=browser --packages=bundle ./node_modules/@fails-components/webtransport/lib/index.node.js --outfile=webtransport-bundle.js`. Add `node:` specifiers where they don't exist in 

```
import { Http3Server } from "./webtransport-bundle.js";
import certificates from "../cert.json" with { type: "json" };

let requests = 0;
const server = new Http3Server({
  port: 8080,
  host: "127.0.0.1",
  secret: certificates[0].secret,
  cert: certificates[0].pem,
  privKey: certificates[0].privateKey,
});

await server.startServer();
await server.ready;

console.info("server started");

const address = server.address();

const sessionStream = server.sessionStream("/");
const sessionReader = sessionStream.getReader();

while (true) {
  const { value: session, done } = await sessionReader.read();
  if (done) {
    break;
  }
  console.log(session.closed);
  let incomingTotalLength = 0;
  let incomingCurrentLength = 0;
  const buffer = new ArrayBuffer(0, { maxByteLength: 4 });
  const view = new DataView(buffer);
  const encoder = new TextEncoder();
  const decoder = new TextDecoder();
  for await (
    const { readable, writable } of session.incomingBidirectionalStreams
  ) {
    const writer = writable.getWriter();
    await readable
      .pipeTo(
        new WritableStream({
          async write(value) {
            if (incomingTotalLength === 0 && incomingCurrentLength === 0) {
              buffer.resize(4);
              for (let i = 0; i < 4; i++) {
                view.setUint8(i, value[i]);
              }
              incomingTotalLength = view.getUint32(0, true);
              console.log(value.length, incomingTotalLength);
              value = value.subarray(4);
            }
            const encoded = encoder.encode(
              decoder.decode(value).toUpperCase(),
            );
            await writer.ready;
            await writer.write(encoded);
            await writer.ready; //.then(() => console.log(`Done writing ${encoded.length} to writable, ${} total bytes written.`));
            incomingCurrentLength += encoded.length;
            console.log(
              `Done writing ${encoded.length} bytes to writable, ${incomingCurrentLength} of ${incomingTotalLength} bytes written.`,
            );
          },
          close() {
            console.log("readable closed");
          },
        }),
      ).then(() => console.log("pipeTo() done"))
      .catch((e) => console.log(e));

    buffer.resize(0);
[...]

PR #366, opened 2025-04-18
Catch empty chunks in streams

martenrichter:


martenrichter:
Closes #365

martenrichter:
Linking https://github.com/w3c/webtransport/issues/643

martenrichter:
Note the discussion about datagrams is not final. I am now dropping empty datagrams, as it would crash the server. WG says I should not drop, but this is not supported on byob reader.

Issue #381, opened 2025-05-16
The logic in stream ID direction detection seems wrong


KershawChang:
The current logic for determining whether a stream is incoming or not seems incorrect to me.
Specifically, the logic:
https://github.com/fails-components/webtransport/blob/347516984ba4f02d505b2a8a7f8664b655b598c9/main/lib/http2/parserbase.js#L118

For example, when streamid = 4n:
 - This is a client-initiated bidirectional stream.
 - If this.isclient === false (i.e., we are the server), this stream should be incoming.
 - However, this code returns false.

As I understand it, stream ID 4 should be treated as an incoming stream on the server side.
Am I missing something?

martenrichter:
The problem seems to be, that I have mixed up server and client initiated streams, see:
https://github.com/fails-components/webtransport/blob/347516984ba4f02d505b2a8a7f8664b655b598c9/main/lib/http2/streamidmanager.js#L299
It is also the other way round. (I looked in the spec, client initiated should be even and server odd).
Expect more problems like this in the HTTP/2 code, as there was no second implementation for testing.
I will try to fix it tomorrow. 
What gives me a bit of a headache is that my Webtransport Over Websocket code also uses this logic, and I am not sure how I can seamlessly transition the users of my lecture web app.

martenrichter:
Ok, I have made a patch.
But I will not merge it, as it is a breaking change in protocol, and I assume that you will find additional issues.
The plan is to merge it and do a second major release once you tell me that it seems to speak a compatible http/2 web transport to Firefox.
Until then, you can use the branch `http2fix` .

Btw. do you know the local testing feature of this package:
https://github.com/w3c/webtransport/issues/622
for http/2, you have to start it with an added 'http2' to the startup command line.

Is there already a http/2 web transport implementation in nightly?

RealTimeCom:
I do not see HTTP/3 `:method` GET in `setRequestCallback()` HTTP function handler, and the server responds with default 404 `file not found`.
Only `:method` CONNECT is handled by `setRequestCallback()` ? there is no way to handle `GET` requests by `Http3Server` ?

Thanks!

martenrichter:
@KershawChang Thanks for the update. I hope it resumes soon....

PR #383, opened 2025-05-17
Fixes errors in the http/2 implementation

martenrichter:
This branch collects fixes for issues in the http/2 implementation brought by Firefox developers to our attention.
It is a breaking change, as streamid semantics change, and will therefore require a semver major release.

Merging will wait until the implementation seems compliant enough to talk to Firefox.


martenrichter:
Fixes #381

PR #387, opened 2025-05-24
Support empty protocol

martenrichter:


martenrichter:
https://github.com/w3c/webtransport/issues/653

Issue #409, opened 2025-09-07
this.socketInt.getSendQueueCount is undefined in transports/http3-quiche/lib/socket.js


guest271314:
I found this by testing several JavaScript runtimes using the `WebTransport` client from this repository, along with the `WebTransport` server from this repository.

Bun throws 
```
terminate called after throwing an instance of 'Napi::Error'
  what():  Invalid state: Controller is already closed
Aborted
``` 
when closing the `WebTransport` (`WebTransportBase` class) client connection

Deno throws this when closing the connection client side, which gives us an object reference

```
terminate called after throwing an instance of 'Napi::Error'
  what():  this.sockterminate called after throwing an instance of 'Napi::Error'
  what():  Invalid state: Controller is already closed
AbortedetInt.getSendQueueCount is not a function
Aborted
```

There are two (2) references to `this.socketInt.getSendQueueCount()` in this code base that I could find, here, where the function is called, yet never defined

https://github.com/fails-components/webtransport/blob/1f42b584321bf4f26f282858b9e40d63db0c03cc/transports/http3-quiche/lib/socket.js#L131-L140

guest271314:
Is that referencing https://nodejs.org/api/dgram.html#socketgetsendqueuecount?

guest271314:
Even when using `node` (`v25.0.0-nightly202507295335c101a9`) Looks like `this.close` is still `false` *after* 

```
closeClient() {
      console.log("closeClient");
      process.nextTick(() => {
        this.cobj.closeClientInt();
        this.socketInt.close();
        this.closed = true;
      });
    }
```

is executed at 

```
sendPacket({ msg, offset, length, port, address }) {
    console.log("sendPacket", this.closed, this.socketInt.getSendQueueCount());
    if (this.closed)
      return true;
    this.socketInt.send(msg, offset, length, port, address, this.packetSendCB);
    const blocked = this.socketInt.getSendQueueCount() > 0;
    this.blocked = this.blocked || blocked;
    return blocked;
  }
```

in `transport-http3-quiche/lib/socket.js`, and ` this.socketInt.getSendQueueCount()` always returns `0`, for input `65507*2`

```
user@debian:~/bin/webtransport-echo-server-bundle$ node --no-warnings wt-client.js
sendPacket false 0
packetSendCB false 0
sendPacket false 0
sendPacket false 0
packetSendCB false 0
packetSendCB false 0
sendPacket false 0
packetSendCB false 0
sendPacket false 0
[...]

PR #412, opened 2025-09-10
If this.socketInt.getSendQueueCount is undefined define function that returns 0; support Deno

guest271314:
Fixes https://github.com/fails-components/webtransport/issues/409

### DIFF
--- a/transports/http3-quiche/lib/clientsocket.js
+++ b/transports/http3-quiche/lib/clientsocket.js
@@ -41,6 +41,10 @@ export class Http3WebTransportClientSocket extends Http3WebTransportSocket {
           ipv6Only: this.forceIpv6
         })
 
+        if (!this.socketInt.getSendQueueCount) {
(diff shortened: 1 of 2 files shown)


martenrichter:
No, I will not do this. A function to probe, if the packet will be droped and that gives information about backpressure on the socket is important for a proper function of the server and client (even if tests seem to work).

martenrichter:
Ok, if bun does this. the implementation will not work reliable. If you have high data transfer rates, the runtime will drop packages, which will have a big negative impact on performance. So it is very very likely that you will run into subtile performance problems or connection issue, where you will have no idea from where they will come. 
So long story short, any runtime not implementing this function properly will not work reliabily.... It is a must have for using libquiche on top of an udp socket. (And yes it will appear as if it works...., but at some point it will not.) So my advice is do not use it with either of the runtimes until they expose the information of the underlying socket.

PR #475, opened 2026-05-24
docs: update README to reflect expanded prebuilt binary support

skrukwa:
The current `README.md` documentation states:
> *"The package provides prebuild binaries for windows, linux and macos for the platform x64..."*

However, looking at the latest GitHub Releases, the CI pipeline is successfully generating and attaching prebuilt `napi` binaries for several other architectures:
- `darwin-arm64`
- `darwin-x64`
- `linux-arm64`
- `linux-x64`
- `win32-ia32`
- `win32-x64`


martenrichter:
As CI/CD is stuck, I am including the fix in https://github.com/fails-components/webtransport/pull/472 with edits. (But also dropping windows ia32 support, as the runners are not supported by github anymore, but still working). Thanks.

Issue #442, opened 2025-12-31
quic_bug_12724_4: Tried to add writer into blocked list when it shouldn't be added


Wain-1:
I sometimes see quic_bug_12724_4 thrown when stress-testing with large (6 MB) writes and not awaiting writer.ready.

Wain-1:
socketjswriter might need to set writeBlocked_ = true

Wain-1:
Note, it doesn't crash, it just adds to the error log.

martenrichter:
> Note, it doesn't crash, it just adds to the error log.

So, what should I see? Just something on stderr?  (Can you post an example?)
P.S: adding an setTimeout to received packages of 40 ms does not trigger anything. (Debugging such errors is hopeless sometime).
It may be, that you have to compile the binary part yourself, if some debug code.

martenrichter:
Ok, I thought it was a quic bug command is, that is terminating (I have confused it with another quic bug command). Then ignore and may be post this issue to https://github.com/[google/quiche](https://github.com/google/quiche). They may want to investigate, if they can reproduce it this way. Currently, it is only showing a programming error, which is catched before it can make damages.

Wain-1:
This is Fable's analysis:


Root-caused it. It's in this repo, not quiche — `writeBlocked_` was the right lead, here's the full chain.

**What happens**

1. `sendPacket` always hands the datagram to `socketInt.send()` first, then returns `getSendQueueCount() > 0` ([socket.js#L120-L133](https://github.com/fails-components/webtransport/blob/bc30fdda8484db7191a13c765ed8690833df440c/transports/http3-quiche/lib/socket.js#L120-L133)). Node only queues into libuv (making that count non-zero) when the kernel synchronously refused the datagram (`uv_udp_try_send` → EAGAIN/EWOULDBLOCK/ENOBUFS) or an earlier refusal is still queued.
2. `SocketJSWriter::WritePacket` turns that into `WriteResult(WRITE_STATUS_BLOCKED, 0)` but never sets `writeBlocked_` ([socketjswriter.cc#L34-L43](https://github.com/fails-components/webtransport/blob/bc30fdda8484db7191a13c765ed8690833df440c/transports/http3-quiche/src/socketjswriter.cc#L34-L43)), so `IsWriteBlocked()` keeps returning false. The `setCanWrite()` in the header has no callers — never had, in any revision.
[...]

PR #492, opened 2026-07-11
Fix adding writer to blocked list

martenrichter:
Co-authored-by: Wain-1


martenrichter:
May fixes:
https://github.com/fails-components/webtransport/issues/442

martenrichter:
Fixes: https://github.com/fails-components/webtransport/issues/442

PR #510, opened 2026-09-06
Remove closed streams from streamObjs

v0id-user:
Every stream a session creates is added to `streamObjs` in `HttpWTSession.onStream`, and the only `removeStreamObj` call runs after a local `writable.abort()`. A stream closed with `writable.close()` and a FIN read to the end leaves the other four Sets but stays in `streamObjs`, with everything it holds, until the session closes.

Opening bidirectional streams, writing one byte each way and closing both sides, heapUsed grows 11.6 KB per stream, linear over 16,000 streams. A snapshot diff shows one `HttpWTStream` per stream per side, retained through `HttpWTSession.streamObjs`.

This change adds `removeStreamObjIfFinished()` to `HttpWTStream`. It calls `removeStreamObj` once both halves are closed and runs from the writable's `close` and `abort` sinks, the readable's `cancel`, the FIN branch of `commitReadBuffer` and `onStreamRecvSignal`.

Two tests, in `bidirectional-streams.spec.js` and `unidirectional-streams.spec.js`, open ten streams on one session, close each, and expect `session.streamObjs.size` to be 0. On master both fail with `expected 10 to equal +0`. With the change they pass and the measurement is flat: 6.4 MB heapUsed after 8,000 streams and after 16,000.

Reproduction, snapshots and DevTools views: https://github.com/v0id-user/fails-components-webtransport-stream-leak


martenrichter:
Thanks for tracking down the issue.
Instead of your changes, I will simplify the whole internal api, which hopefully will prevent similar errors in the future.
See https://github.com/fails-components/webtransport/pull/511 .
A new release will follow after CI passes.

v0id-user:
Thanks @martenrichter for the quick turnaround. Tested 1.6.8 with the same script, no growth now over 16k streams.

